PR #13, opened 2023-08-30
fix: bad engine version

xhayper:
## Changes

Fix bad engine version caused by improperly configured `engines` property in `package.json`

`@octokit/rest` 20.0.1 requires node version higher than 18, which conflicts with creator-docs's requirement of node version higher than 16.

This pull request also update the lockfile.

## Checks


xhayper:
That is understandable!

PR #34, opened 2023-08-30
Update metatables.md

metatablecat:
## Changes

* ~~Remove `__gc` [[1]](https://devforum.roblox.com/t/luau-recap-may-2020/581996) (only one I could find that references the removal)~~
* Clarify that equality methods require the same metamethod, not metatable [[2]](https://github.com/Roblox/luau/issues/386#issuecomment-1048259959)
* Add initial `__iter` documentation (probably needs further clarification [[3]](https://luau-lang.org/syntax#generalized-iteration)
* 
## Checks


content/en-us/luau/metatables.md
```diff
@@ -118,19 +118,19 @@ Here's the list of available metamethods:
 		  <td>__mode</td>
 		  <td>Used in weak tables, declaring whether the keys and/or values of a table are weak. **Note:** References to Roblox instances are never weak. Tables that hold such references will never be garbage collected.</td>
 		</tr>
-	  <tr>
-		  <td>__gc(table)</td>
-		  <td>Fired when the table is garbage-collected. **Note:** On Roblox, this metamethod is disabled.</td>
-		</tr>
```
aetter, line 124: Hey @metatablecat, I think it's still worth leaving this in to give people some context in the (rare) case that they were expecting to see it. Maybe something like, `This metamethod is present in Lua, but removed in Roblox. Declaring it does nothing.` Thoughts?

content/en-us/luau/metatables.md
```diff
@@ -118,19 +118,19 @@ Here's the list of available metamethods:
...
 	  <tr>
 		  <td>__len(table)</td>
 		  <td>Fired when the # length operator is used on the Object.</td>
 		</tr>
+	  <tr>
+		  <td>__iter(table)
+		  <td>Used to denote a custom iterator when using generalized iteration.</td>
+		  </td>
```
aetter, line 128: Awesome, thanks for adding this. Agreed that we should add a note to https://create.roblox.com/docs/luau/control-structures#for-loops as well, but that doesn't need to be done in this PR.

content/en-us/luau/metatables.md
metatablecat: This seems like it could be an interesting choice, `__gc` is removed for security reasons so I feel like it would be better to lean it towards that as the reason

aetter:
Thank you @metatablecat!

PR #41, opened 2023-08-30
refactor: use `ipairs` for array instead of `pairs`

xhayper:
## Changes

Replace all instance of `pairs` that was used with an array with `ipairs` as `ipairs` is much faster than `pairs` when used with an array

## Checks

### DIFF
--- a/content/en-us/animation/using.md
+++ b/content/en-us/animation/using.md
@@ -106,7 +106,7 @@ climbing, swimming, and jumping. You can replace these [default animations](#def
    	local animator = humanoid:WaitForChild("Animator")
 
    	-- Stop all animation tracks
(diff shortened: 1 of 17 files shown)


christopher-buss:
To chime in: I don't think this is a good change. Luau's generalized iterator should be encouraged over the use of ipairs (https://luau-lang.org/syntax#generalized-iteration).

Kampfkarren:
Yes, general recommendation is not using pairs or ipairs anymore.

aetter:
Hey @xhayper, thanks a bunch for sending this over. My take here is that even if there's a performance uplift from the `ipair` change, we'd rather leave things as-is and avoid the additional testing until we have time to do a proper refactor with more Luau-specific syntax.

PR #45, opened 2023-08-30
Bump the memory requirements to 4GB from 2GB

idontyboi:
##### First time attempting to contribute to Open Source, hope it goes alright...

**Short Summary:** Studio has outgrown my existing 2GB computer, making Studio unusable, with it being total opposite of a smooth experience. The documentation still says it should be smooth with that low memory, although recent changes have made it anything but that. So, bump up the memory requirements.

**Long Summary:**
This won't be taken lightly, so here's how it is in detail. My main computer is dated 2009/2010, low-end of it's time, with the following specifications:
- LGA775 Pentium E6500 Dual-Core @ 2.93 GHz
- 2.00 GB of system memory, 1-channel, DDR3
- Intel G41 Express Chipset (DirectX 10)
- Windows 7 x64
- A 7200RPM HDD, ~80GB free
- 8 M**b**/s internet connection (1 Mb/s upload)

Looks like it matches almost exactly the minimum requirements. The documentation says that Studio should run smoothly. How smoothly does it run in this configuration? Well...
Calling it not smooth would be a understatement: it's, at best, *glacier.*
- It takes 3 minutes to start up with a full minute and a half of it being frozen [not responding] on the login screen.
- The home page alone takes over 780 MB, a simple baseplate takes over 1.2 GB and pressing Play on a simple baseplate takes 1.8 GB of memory (commit). And each stage takes atleast a minute, with the disk utilization pegged at 100% HDD usage. 
- Anything more simple than a baseplate is borderline unusable, with simplest games taking far longer to make. It also limits the amount of developer tools you can use. Don't even dare to work with Rojo, because Studio will time out when starting CoreScripts.
In 2019, 2020 and early 2021 it was completely usable. It took a sudden turn after that, and I keep a copy of a version of Studio just before the sudden turn.

While you could delete built-in plugins and scoop up a dozen FFlags by your own to reduce memory usage to make it usable, it only gets usable at simple games and you still have to get a friend to press Play and report errors 24/7 in order to avoid the thrashing.
And yes, I am putting my **own** PC out of the minimum system requirements.

There is a bug report that [complains about memory deficiencies in plugins](https://devforum.roblox.com/t/unused-or-internal-built-in-plugins-are-always-loaded-causing-excess-memory-usage/2466797), which was not answered by any Roblox-official triager and is assumed to be not looked at.
Fixing this almost certainly requires communication across 63 teams (1 for each plugin). Since it's unlikely to be fixed, just make it clear that Studio will be laggy by bumping the memory requirement to 4GB.
## Checks


aetter:
Hi @idontyboi, thanks a bunch for bringing this up. I'm trying to chase it down internally. Like you, I can't imagine running on 2 GB of RAM with a modern OS is a good experience.

idontyboi:
Be glad I didn't test this under Windows 10 :)
Thinking if I should bump this up further to 8GB. It got *a lot* worse withnin the past month and a half, so not sure that 4GB is enough for a good Studio experience based on measurements.

## Measurements
This table was measured using the same horrible specs up above. Latest Studio version 0.592, without any changes or modifications to make it usable on this old PC. Just the web browser [Chrome 109.0.5414.149] was open, with one tab active.
Commit Charge: the amount of virtual memory allocated. This is good to measure allocation above RAM size (Win7 Task Manager shows it as KB)
Usable Time: The amount of time it takes for the feature to actually become usable.

Action | Commit Charge (KB) | Usable Time | Timeout or Comments
--- | --- | --- | ---
Open Studio | 966704 | 1:20 | Already 50%
Open Baseplate | 1444472 | 1:30 | The regular one, not classic.
Play Solo | 2117924 | 2:10 | And Studio is using more Commit than RAM
Stop Play | 1957452 | 2:00 | Play Solo allocated 673 MB, but only 157 MB freed up, hmm
Close Baseplate | 1641872 | 1:15 | Open Baseplate allocated 466 MB, but only 308 MB freed up, hmm?
Re-Open Studio | N/A | N/A | To properly free up, I have to re-open Studio
Open Modern City | 2534428 | 11:35 | Took so long that LoadingScript times out at line 11. Not sure about negative impacts.
Play Solo | 3312860 | 12:45 | Took so long that StarterScript times out, resulting in the Roblox base UI not appearing at all.
Re-Open Studio | N/A | N/A | To properly free up, I have to re-open Studio
Open Baseplate | N/A | N/A | And open baseplate.
Local Server 2plr | 5740040 | 79:00 | Creates 4 Studio instances. Individual Instance mem 1474160 + 1458820 + 1435976 + 1371084. No timeouts, suprisingly. **And yes, I did wait over an hour for this.**

Investigation shows that built-in plugins are a large cause and I think it is underlooked. Not only they are loaded at all times, but each one copies a duplicate copy of a ton of CorePackages instead of pointing directly *at* it, equalivent to 65 duplicate copies of various node_modules. And these packages are *not light.* Those plugins have to be loaded *two* times on Play Solo, if not *three* times if switching to Server context.

*Just a 2-player Local Server on a empty baseplate takes up more commit charge than 4GB and can not fit in all in 4GB RAM as-is.*
*Most users running Studio will likely multitask: They have a web browser to get documentation, Discord/Guilded app [**Electron**] to join a developer community. Throw in Windows 11 in the mix, and you now got yourself a strapped system where Windows has to constantly reduce and thrash memory.*

Windows 11 can't possibly fit in all this under just 4 gigabytes without a lot of lag. If running on a SSD, it certainly won't be *that* laggy, but will consume a lot of write cycles and shorten SSD lifespan. With all this, should I bump it to 8GB or is it fine as-is?

aetter:
I think 4 GB is still the right call for the time being -- if people are running Windows 7 and have a handle on their background applications, it might be fine. I'm sure we'll continue bumping it up a bit over the years. 😄 

I'm going to merge and then make some additional changes on my end around OS versions. Thanks again for bringing this up and for the wealth of research!

aetter (approved):

PR #81, opened 2023-08-31
Add Generalized Iteration to control-structures

metatablecat:
## Changes

Adds basic documentation on generalized iteration syntax `for ... in {} do` [[1]](https://luau-lang.org/syntax#generalized-iteration)

## Checks


content/en-us/luau/control-structures.md
```diff
@@ -166,6 +166,32 @@ The generic `for` loop iterates over items in a collection rather than a sequenc
 
 For loops need a function, or iterator, to iterate over different types of collections. The global `ipairs()` returns an iterator for arrays, and the global `pairs()` returns an iterator for dictionaries. The `Library.string` library provides `Library.string.gmatch()` to iterate over strings.
 
+### Generalized Iteration
+
+In Luau, it is possible to iterate over a table by using the `in` keyword directly on the table, instead of using an iterator function such as `ipairs()`.
```
aetter, line 171:
```suggestion
In Luau, you can iterate over a table using the `in` keyword directly on the table, instead of using an iterator function such as `ipairs()`:
```

content/en-us/luau/control-structures.md
```diff
@@ -166,6 +166,32 @@ The generic `for` loop iterates over items in a collection rather than a sequenc
...
+
+```lua
+for i, v in {1, 2, 3, 4, 5} do
+	print(i, v)
+end
+```
+
+The `__iter` metamethod can be used to create a custom iterator function, that can be called when using generalized iteration.
```
aetter, line 179:
```suggestion
Generalized iteration also lets you use the `__iter` metamethod to create a custom iterator function:
```

content/en-us/luau/control-structures.md
```diff
@@ -166,6 +166,32 @@ The generic `for` loop iterates over items in a collection rather than a sequenc
...
+	__iter = function(o)
+		return next, o.items 
+	end
+})
+
+for k, v in obj do
+  assert(k * k == v)
+end
+```
+
```
aetter, line 194: Hey @metatablecat, I know this example is straight from the Luau documentation, but I think something like this is a bit easier to understand (especially for someone new to metatables):

```lua
local table = {1, 4, 9}

local metatable = { 
	__iter = function(self)
		print("This prints one time.")
		return next, self
	end
}

setmetatable(table, metatable)

for i, v in table do
	print(i, v)
end
```

Any thoughts? Am I missing something important by not using the `obj` `o` framing?

metatablecat: I feel like an example should be used to showcase that the returned function has the same arguments and returns as the `next` function.

The type of __iter is (assuming `self` refers to the table) is

`__iter<T>(self) -> ((T, lastIndex: any) -> ...any), T)`

aetter: Yeah, I've been struggling to find something that isn't completely contrived -- the good use case here seems super hard to define. This sample feels... pretty poor, but maybe it demonstrates the concept better than just using `next`. All it does is go last to first instead of first to last. Let me know what you think:

```lua
local myTable = {0, 2, 4, 6, 8, 10, 12, 14, 16}

myMetatable = { 
	__iter = function(self)
		local i = #self
		local firstRun = true
		return function()
			local value
			if firstRun then
				firstRun = false
				return i, self[i]
			else
				i -= 1
				if i > 0 then
					return i, self[i]
				end
			end
		end
	end
}

setmetatable(myTable, myMetatable)

for i, v in myTable do
	print(i, v)
end
```

content/en-us/luau/control-structures.md
```diff
@@ -166,6 +166,32 @@ The generic `for` loop iterates over items in a collection rather than a sequenc
...
+setmetatable(obj, {
+	__iter = function(o)
+		return next, o.items 
+	end
+})
+
+for k, v in obj do
+  assert(k * k == v)
+end
```
aetter, line 192:
```suggestion
local myTable = {1, 2, 3, 4, 5}

myMetatable = { 
	__iter = function(self)
		local i = #self
		local firstRun = true
		return function()
			if firstRun then
				firstRun = false
				return i, self[i]
			else
				i -= 1
				if i > 0 then
					return i, self[i]
				end
			end
		end
	end
}

setmetatable(myTable, myMetatable)

for i, v in myTable do
	print(i, v)
end
```

aetter (approved):

aetter:
Thanks a bunch, @metatablecat! I'm merging this, but we can always iterate on the code sample in new PRs you think there's a better way to convey the point.

PR #8, opened 2023-08-29
security-tactics improvements and additions

xonae:
## Changes

I made the following changes to improve pre-existing code:
- The `typeValidator` tuple type validator in the Remote Runtime Type Validation was created on-the-fly every single time a request was made from the client. If something can be made once and reused infinitely, it should be!
- Instead of throwing an error on the server when the `typeValidator` type check fails, which could trigger analytics requests and other heavy error handling logic (among other things) on an event that is *not* rate limited, it silently fails and returns that information to the client to receive, display and act on accordingly. In this example it isn't a fatal error that should be logged in that way.
- Adjusted some of the "highlight" data in the Lua code blocks. I presume these are used to highlight important parts of the code samples, though I've never seen this myself on the creator docs. Regardless, they should be more accurate for the code snippets now.

I added the following additions to the article:
- Added an example of "instance spoofing" under Data Validation that ties in with the existing "In-Experience Shop" example. This is an incredibly common and easily overlooked exploit that can be taken advantage of in many games, and definitely deserves a spot in this article.
- Added an in-depth list of various attacks that can prevent DataStore saving from malicious client input under the right circumstances, and how to prevent them. The listed attacks can be detrimental to games and could result in item duplication among other game-breaking actions based on the type of game.

I've given everything a good look over based on the style guide and rewritten a lot from my initial draft. Please correct anything that I might've missed. Cheers!

## Checks

By submitting your pull request for review, you agree to the following:

- [X] This contribution was created in whole or in part by me, and I have the right to submit it under the terms of this repository's open source licenses.
- [X] I understand and agree that this contribution and a record of it are public, maintained indefinitely, and may be redistributed under the terms of this repository's open source licenses.
- [X] To the best of my knowledge, all proposed changes are accurate.

PR #203, opened 2023-10-25
Expand and correct documentation on data store string sanitisation.

TheGreatSageEqualToHeaven:
The documentation that explains how to sanitise strings is problematic and does not properly solve the problem it claims it solves. Data stores can save valid UTF-8 while the provided fix which was borrowed from my [github gist](https://gist.github.com/TheGreatSageEqualToHeaven/e0e1dc2698307c93f6013b9825705899#patch-1) only allows ascii characters to be passed, it is not a real fix and should not be recommended.

This documentation should also be further updated if my [feature request](https://devforum.roblox.com/t/datastoreservice-method-to-verify-if-a-value-is-saveable-bring-more-attention-to-malicious-strings/2658551) were to be accepted and introduce a proper solution. 

image

Edit:

Luau now has [stricter UTF-8 validation](https://github.com/luau-lang/rfcs/blob/master/docs/behavior-stricter-utf8-library.md) as of 2 weeks ago so we can replace the UTF-8 verification function with `utf8.len` and expect it to work properly.


Jonathan23603:
Me gusta mucho

TheGreatSageEqualToHeaven:
Ping

TheGreatSageEqualToHeaven:
Any updates on this?

xonae:
Hey, just poking my head in here. Great pull request! I was the author of the [pull request](https://github.com/Roblox/creator-docs/pull/8) that features the code snippet mentioned. The string matching regex was not borrowed from your gist as you've claimed, it achieves the same goal but is completely different and something I put together myself. I can understand the confusion though!

While this new function you provided does a much better job than basic string matching I think it's a bit too bulky to provide as-is on the page. Perhaps instead consider attaching it as a gist or repository similar to how the "t" library is linked under Remote Runtime Type Validation?

But I agree that this should probably be a built-in method that's referenced clearly in the documentation for SetAsync/UpdateAsync/etc. methods that would benefit from its usage. I hope your [feature request](https://devforum.roblox.com/t/datastoreservice-method-to-verify-if-a-value-is-saveable-bring-more-attention-to-malicious-strings/2658551) gets added!

TheGreatSageEqualToHeaven:
> While this new function you provided does a much better job than basic string matching I think it's a bit too bulky to provide as-is on the page. Perhaps instead consider attaching it as a gist or repository similar to how the "t" library is linked under Remote Runtime Type Validation?

Providing bad advice for the sake of simplicity is not a solution. Having accuracy and allowing international inputs is a requirement for games when they're not just shipped to English speaking countries, this also affects borrowed words like naïve. The fact that it is bulky goes to show that Roblox should have a proper solution for this rather than forcing us developers to have to create these solutions. [Stricter UTF8 validation](https://github.com/luau-lang/rfcs/blob/master/docs/behavior-stricter-utf8-library.md) was accepted into Luau after a lot of discussion so we have a proper solution now with `utf8.len` (properly this time though!) so I'll update this pr once I am free to use that instead, although like how Arseny said there should be a function for verifying the validity.

TheGreatSageEqualToHeaven:
Ping, any updates on this?

TheGreatSageEqualToHeaven:
@ahyatt-rbx Updated the documentation to use `utf8.len` since [stricter UTF-8 validation](https://github.com/luau-lang/rfcs/blob/master/docs/behavior-stricter-utf8-library.md) was added. This should be good to merge now.

content/en-us/scripting/security/security-tactics.md
```diff
@@ -209,7 +209,7 @@ Ensure that any actions performed through a `Class.RemoteEvent` or `Class.Remote
...
 - `Class.DataStore|DataStores` have [data limits](../../cloud-services/datastores.md#data-limits). Strings of arbitrary length should be checked and/or capped to avoid this, alongside ensuring limitless arbitrary keys cannot be added to tables by the client.
 - Table indices cannot be `NaN` or `nil`. Iterate over all tables passed by the client and verify all indices are valid.
-- `Class.DataStore|DataStores` can only accept valid UTF-8 characters. Sanitize all strings provided by the client to ensure only bytes 1-127 are used. By using `Library.string.find()` with the pattern `"[%z\128-\255]"`, an operation can be aborted if any invalid characters are found in a string.
+- `Class.DataStore|DataStores` can only accept valid UTF-8 characters. Sanitize all strings provided by the client to ensure they are valid using `utf8.len`. Invalid UTF-8 strings can also be present in tables as keys and values. `utf8.len` will return the length of a string treating unicode characters as a single character, if an invalid UTF-8 character is encountered it will return `nil` and the position of the invalid character.
```
IgnisRBX, line 212:
```suggestion
- `Class.DataStore|DataStores` can only accept valid UTF-8 characters, so you should sanitize all strings provided by the client via `Library.utf8|utf8.len()` to ensure they are valid. `Library.utf8|utf8.len()` will return the length of a string, treating unicode characters as a single character; if an invalid UTF-8 character is encountered it will return `nil` and the position of the invalid character. Note that invalid UTF-8 strings can also be present in tables as keys and values.
```

IgnisRBX (approved):

PR #249, opened 2023-11-08
Fix method name for automatic network ownership on BasePart

a-lil-birb:
## Changes

In the description about setting the network ownership behavior back to automatic, link the method **Class.BasePart:SetNetworkOwnershipAuto()** instead of **Class.BasePart:SetNetworkOwner()**. The latter method with nil parameter will set network ownership to the server rather than the intended automatic behavior.

**SetNetworkOwnershipAuto()** is also the method used in the code snippet below it, so it seems like the original description was a typo.

## Checks

- [✔] This contribution was created in whole or in part by me, and I have the right to submit it under the terms of this repository's open source licenses.
- [✔] I understand and agree that this contribution and a record of it are public, maintained indefinitely, and may be redistributed under the terms of this repository's open source licenses.
- [✔] To the best of my knowledge, all proposed changes are accurate.


Katrist:
Instead of adding an emoji for the checklist, you can use an "X".

Example: [X]

Good PR though.

aetter (approved):

aetter:
Thank you @a-lil-birb!

PR #287, opened 2023-11-19
Update LuaGlobals.yaml - Markdown Correction

PIaceholders:
Markdown correction under global "setfenv"

I believe there was a small spelling mistake while typing the markdown for "setfenv", I was skimming over the Lua Global's documentation and just happened to see it.

Right now it looks like this: If `f` is ` 0``, then `setfenv()`
After the update it should like look like: If `f` is `0`, then `setfenv()`


aetter (approved):

aetter:
Hey @PIaceholders, this looks great, thanks a bunch. I need you to check the three boxes that I've added back to your post before we merge, though.

IgnisRBX (approved):

PR #301, opened 2023-11-22
Clarify behavior of GlobalDataStore.UpdateAsync.

Anaminus:
## Changes

Clarify that, when UpdateAsync invokes the callback again, the result of the previous call is discarded.

Clarify that the callback returning nil only affects the current game server.

## Checks


content/en-us/reference/engine/classes/GlobalDataStore.yaml
```diff
@@ -346,6 +347,9 @@ methods:
         existing metadata is being changed; otherwise all existing metadata will
         be cleared.
 
+      If the callback returns `nil` instead, then the current game server will
```
IgnisRBX, line 350:
```suggestion
      If the callback returns `nil` instead, the current server will
```

PR #305, opened 2023-11-26
Fix imprecise return types for methods of TextService.

Anto-fire:
Fixed incorrect / unprecise return types.

## Changes

Changed the return types of the two methods from Instance to TextFilterTranslatedResult and TextFilterResult respectively.

## Checks


IgnisRBX:
Hello, thanks for the suggestion here. Just FYI, only summaries, descriptions, and deprecation messages can be changed for classes and enums through this workflow; all other changes will be automatically overwritten from the source. I have requested that these return types be updated at the source though, so once that is complete, they should update to something more specific than "Instance".

PR #341, opened 2023-12-11
Correct description of global unpack's parameters

dv-extrarius:
Correct the description of the unpack parameters.

## Changes

The description of `unpack`'s `j` parameter was wrong. It is not a count, it is the index of the last element to `unpack`. Updated the description of the `i` parameter to match the new wording of the `j` parameter.

## Checks

### DIFF
--- a/content/en-us/reference/engine/globals/LuaGlobals.yaml
+++ b/content/en-us/reference/engine/globals/LuaGlobals.yaml
@@ -686,7 +686,7 @@ functions:
         type: int
         default: '#list'
         summary: |
-          The index of the last element to unpack.
+          The number of elements to unpack.
     returns:
       - type: Variant
         summary: ''
--- a/content/en-us/reference/engine/globals/LuaGlobals.yaml
+++ b/content/en-us/reference/engine/globals/LuaGlobals.yaml
@@ -55,7 +55,7 @@ functions:
       - type: Variant
         summary: ''
     tags:
-    code_samples:
+    codesamples:
   - name: collectgarbage
     summary: |
       Performs the specified operation of the garbage collector.
@@ -78,7 +78,7 @@ functions:
       - type: Variant
         summary: ''
     tags:
-    code_samples:
+    codesamples:
   - name: error
     summary: |
       Halts thread execution and throws an error.
@@ -614,7 +614,7 @@ functions:
         type: int
         default: 10
         summary: |
-          The numerical base to convert _arg_ into.
+          The numerical base to convert `arg` into.
     returns:
       - type: Variant
         summary: ''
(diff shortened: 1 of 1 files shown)


content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -681,12 +681,12 @@ functions:
...
         default: 1
         summary: |
-          The starting index to unpack from.
+          The index of the first element to unpack.
       - name: j
         type: int
         default: '#list'
         summary: |
-          The number of elements to unpack.
+          The index of the last element to unpack.
```
IgnisRBX, line 689:
```suggestion
          The number of elements to unpack.
```

IgnisRBX (dismissed):

IgnisRBX:
I reverted the suggestion of the "j" parameter, which isn't the "index of the last element to unpack", but rather the number of indices to unpack (default to the total number of indices) starting from the given index number "i" (default of the first index).

content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -55,7 +55,7 @@ functions:
       - type: Variant
         summary: ''
     tags:
-    code_samples:
+    codesamples:
```
IgnisRBX, line 58:
```suggestion
    code_samples:
```

content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -78,7 +78,7 @@ functions:
...
```
IgnisRBX, line 81:
```suggestion
    code_samples:
```

IgnisRBX (approved):

dv-extrarius:
> I reverted the suggestion of the "j" parameter, which isn't the "index of the last element to unpack", but rather the number of indices to unpack (default to the total number of indices) starting from the given index number "i" (default of the first index).

That's not how it functioned for me in roblox. I had an array and wanted slices so I tried with i=1, 11, 21, etc and j=10 and only the first slice had elements. When I made j=10, 20, 30, etc it functioned as I wanted, slicing 10 elements at a time

Hexcede:
> rather the number of indices to unpack

This is incorrect. Unpack takes two indices, the second argument is not the count but rather the last index. The second argument does default to the length of the input table, but this is because the last index of the table matches the length, not because the argument is a count. This is confirmed by both the [Lua 5.1](https://www.lua.org/manual/5.1/manual.html#pdf-unpack) and [Lua 5.4](https://www.lua.org/manual/5.4/manual.html#pdf-table.unpack) documentation as well as through empirical testing.

getyofuckindogback:
> rather the number of indices to unpack (default to the total number of indices)

I can see how the misunderstanding with the unpack function can be seen here. While yes, parameter `j` is defaulted to total number of elements in a table, it's defaulted to that because you're grabbing from start of `i` and ending at `j`. `j` is wherever you want unpacking to stop. `table.unpack`/`unpack` is basically like javascript's slice prototype function.

https://github.com/luau-lang/luau/blob/e76802f2ce698ca090a793b24c07e336b21ade9f/VM/src/ltablib.cpp#L269C18-L269C18
Dosen't explain much but this is the pushed global cfunction of `unpack`.

PR #375, opened 2024-01-10
Add note about Workspace.PersistentLoaded

legendoj1:
## Changes

Added an important note to the documentation for Workspace.PersistentLoaded about it only being accessible on the server, as mentioned in its release announcement: https://devforum.roblox.com/t/new-improvements-to-streaming-enabled/2185535

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/Workspace.yaml
+++ b/content/en-us/reference/engine/classes/Workspace.yaml
@@ -1232,6 +1232,8 @@ events:
       Note that experience loading happens before persistent loading, and firing
       of the `Class.DataModel.Loaded` event does not indicate that all
       persistent models are present.
+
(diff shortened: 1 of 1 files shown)


IgnisRBX:
Hi there, thanks for the suggestion, but PersistentLoaded can actually be used in a LocalScript (although that's not the typical usage for it). Because of the loading order for persistent models, they are most likely already loaded by the time your LocalScript runs, so the event will not detect them loading (because that happened previous to that time). To test, you can place the LocalScript inside ReplicatedFirst and the event will fire.

IgnisRBX:
We'll update the forum post as well, since it was unclear. It's the properties and functions that are server-side, but the note implied that it was all (our bad). Thanks for flagging this!

PR #382, opened 2024-01-16
Add markdown escape to TextChatService.yaml

mvyasu:
## Changes

The last pull request for this problem fixed most of the problem, but it looks like the bar character in `Class.TextChannel | TextChannels` is now being used for the markdown table. 

IncorrectMarkdown

This is the reason why this pull request adds an escape character before the bar. It is unknown if this will break the automatic link:

`Class.TextChannel \| TextChannels`

I tried to see what was done in such a scenario on other pages, but I was unable to find a similar scenario. The closest I found was this table on [Datatype.Vector3](https://create.roblox.com/docs/reference/engine/datatypes/Vector3) However, it turned out that table just used pure html tags for its table and it doesn't override the default text link. 

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/TextChatService.yaml
+++ b/content/en-us/reference/engine/classes/TextChatService.yaml
@@ -68,14 +68,14 @@ properties:
       created and put in a `Class.Folder` named **TextChatCommands** inside
       `Class.TextChatService`:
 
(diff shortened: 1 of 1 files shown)


IgnisRBX (approved):

IgnisRBX:
Removing the spaces around the " \| " should make the link form correctly (I did so). Thanks for the correction!

PR #413, opened 2024-02-08
Corrected :Clone() return type

michaeldougal:
## Changes

`Instance:Clone()` will return `nil` if it has `Instance.Archivable` set to `false`, so the correct type is `Instance?`

Edit: Since we apparently cannot directly edit type fields, I added a note to the description that describes this particular case.

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/Instance.yaml
+++ b/content/en-us/reference/engine/classes/Instance.yaml
@@ -348,31 +348,24 @@ methods:
     thread_safety: Unsafe
   - name: Instance:Clone
     summary: |
-      Create a copy of an object and all its descendants, ignoring objects that
-      are not `Class.Instance.Archivable|Archivable`.
+      Create a copy of an instance and all its descendants, ignoring instances
+      that are not `Class.Instance.Archivable|Archivable`.
     description: |
-      **Clone** creates a copy of an object and all of its descendants, ignoring
-      all objects that are not `Class.Instance.Archivable|Archivable`. The copy
-      of the root object is returned by this function and its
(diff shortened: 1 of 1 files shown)


content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -352,7 +352,9 @@ methods:
       **Clone** creates a copy of an object and all of its descendants, ignoring
       all objects that are not `Class.Instance.Archivable|Archivable`. The copy
       of the root object is returned by this function and its
-      `Class.Instance.Parent|Parent` is set to nil.
+      `Class.Instance.Parent|Parent` is set to nil. Note that if the object
+      itself has `Class.Instance.Archivable|Archivable` set to false, this
+      function will return nil.
```
IgnisRBX, line 357:
```suggestion
      `Class.Instance.Parent|Parent` is set to `nil`. Note that if the object
      itself has `Class.Instance.Archivable|Archivable` set to `false`, this
      function will return `nil`.
```

IgnisRBX (approved):

IgnisRBX:
I made a few more clarifications to the Archivable property as well; please confirm that this is clearer and I'll merge the PR. Thanks for the contribution!

michaeldougal:
Nice, that does seem much clearer and more accurate to me. I would just point out that in the description for :Clone() the word 'object' is used to refer to the instance in question, while the description for Archivable uses 'instance'. Not a huge deal, but thought I'd mention it for consistency's sake.

IgnisRBX (approved):

PR #418, opened 2024-02-13
Add * format specifier.

metatablecat:
## Changes

Adds the `*` format specifier, which is basically just `%s` but accepts anything, uses the `tostring` result of the object. ({} for example, becomes `table: 0x123456789abcdef` 

(Couldn't find a source on it but can be tested by running this code in Luau)
```lua
local fmt = "%*"
local t_mt = setmetatable({}, {__tostring = function() return "cat" end})

print(fmt:format({})) --> table: 0x123456789abcdef
print(fmt:format(t_mt)) --> cat
print(fmt:format(print)) --> function: 0x123456789abcdef
```

## Checks


content/en-us/reference/engine/libraries/string.yaml
```diff
@@ -187,6 +187,12 @@ functions:
             <td>Hexadecimal representation.</td>
             <td>7fa<br>7FA</td>
           </tr>
+          <tr>
+            <td><code>*</code></td>
+            <td>any</td>
+            <td>Equivalent to <code>s</code> but accepts any variable by converting it to a string using <code>tostring</code>.</td>
```
IgnisRBX, line 193:
```suggestion
            <td>Equivalent to <code>s</code> but accepts any variable by converting it to a string using <code>Global.LuaGlobals.tostring()</code>.</td>
```

IgnisRBX (approved):

PR #438, opened 2024-02-24
more precise details for TextChatService.MessageReceived

GFinkGH:
## Changes

I've clarified some slightly ambiguous technical details for the `TextChatService.MessageReceived` event. Before, it wasn't entirely clear if the event was only fired on the client because saying "fired exclusively on the client when [...]" or "fired exclusively on the client under two circumstances [...]" suggests that there are cases where it may be fired in contexts other than the client under different circumstances. Rather than finesse the sentence to make it clear that it's only ever used on the client, I just added a new sentence later in the same way [`TextChannel::SendAsync`](https://create.roblox.com/docs/en-us/reference/engine/classes/TextChannel#SendAsync) does. I also added information about how `TextChannel.ShouldDeliverCallback` affects `TextChatService.MessageReceived`.

This is my first time ever using GitHub. I was told I could edit Creator Docs pages through here on the DevForum: https://devforum.roblox.com/t/_/2706053/7

## Checks


GFinkGH:
Not sure what the `"source.fixAll.markdownlint": "explicit"` change is, I don't remember messing with that. I'm new, forgive me 🙈

.vscode/settings.json
```diff
@@ -22,7 +22,7 @@
   },
   "[markdown]": {
     "editor.codeActionsOnSave": {
-      "source.fixAll.markdownlint": true
+      "source.fixAll.markdownlint": "explicit"
```
IgnisRBX, line 25:
```suggestion
      "source.fixAll.markdownlint": true
```

content/en-us/reference/engine/classes/TextChatService.yaml
```diff
@@ -283,13 +283,13 @@ events:
     thread_safety: Unsafe
   - name: TextChatService.MessageReceived
     summary: |
-      Fires exclusively on the client in two circumstances: when the client
-      receives a `Class.TextChannel:SendAsync()` response from the server, or
-      when `Class.TextChannel:DisplaySystemMessage()` is invoked on the client.
+      Fires when `Class.TextChannel:DisplaySystemMessage()` is invoked on the client, or when the client receives a valid `Class.TextChannel:SendAsync()` response from the server.
```
IgnisRBX, line 286:
```suggestion
      Fires when `Class.TextChannel:DisplaySystemMessage()` is invoked on the
      client, or when the client receives a valid `Class.TextChannel:SendAsync()`
      response from the server.
```

content/en-us/reference/engine/classes/TextChatService.yaml
```diff
@@ -283,13 +283,18 @@ events:
...
-      receives a `Class.TextChannel:SendAsync()` response from the server, or
-      when `Class.TextChannel:DisplaySystemMessage()` is invoked on the client.
+      Fires when `Class.TextChannel:DisplaySystemMessage()` is invoked on the
+      client, or when the client receives a valid `Class.TextChannel:SendAsync()`
+      response from the server.
+      
+      If the server's `Class.TextChannel.ShouldDeliverCallback` property is bound
+      and returns `false`, the client will not fire `Class.TextChatService.MessageReceived`.
+      
+      This event is only fired on the client.
```
IgnisRBX, line 297:
```suggestion
      Fires when `Class.TextChannel:DisplaySystemMessage()` is invoked on the
      client, or when the client receives a valid `Class.TextChannel:SendAsync()`
      response from the server. This event is only fired on the client.
      
      If the server's `Class.TextChannel.ShouldDeliverCallback` property is bound
      and returns `false`, the client will not fire `Class.TextChatService.MessageReceived`.
      
```

content/en-us/reference/engine/classes/TextChatService.yaml
```diff
@@ -283,13 +283,17 @@ events:
...
-      Fires exclusively on the client in two circumstances: when the client
-      receives a `Class.TextChannel:SendAsync()` response from the server, or
-      when `Class.TextChannel:DisplaySystemMessage()` is invoked on the client.
+      Fires when `Class.TextChannel:DisplaySystemMessage()` is invoked on the
+      client, or when the client receives a valid `Class.TextChannel:SendAsync()`
+      response from the server. This event is only fired on the client.
+      
+      If the server's `Class.TextChannel.ShouldDeliverCallback` property is bound
+      and returns `false`, the client will not fire `Class.TextChatService.MessageReceived`.
+
```
IgnisRBX, line 296:
```suggestion
```

GFinkGH:
> I made a few tiny edits to your suggestions... how does it look now?
> 
> I think TextChannel.MessageReceived may also need some edits... could you please suggest those too? Thanks!

Testing shows it follows the same rules as TextChatService.MessageReceived, so I assume it would just be copy-and-paste (also those pages currently have the same text for this event).

Repro is attached with relevant comments; GitHub doesn't support .rbxm files so it's zipped here: [MessageReceived repro.zip](https://github.com/Roblox/creator-docs/files/14411471/MessageReceived.repro.zip)

GFinkGH:
Are the oddly placed linefeeds intentional? Maybe it looks different in the final version but I'm wondering if there's a guideline for restricting the number of characters per line. It looked weird in the editor so I deleted some of them.

Also, should it be mentioned in the docs that this event works the same with both the TextChatService and TextChannel instances? Maybe copy-pasting isn't the only thing that should be done

content/en-us/reference/engine/classes/TextChannel.yaml
```diff
@@ -118,23 +118,22 @@ methods:
...
       and the text of the message (with `Class.TextChatMessage.Text`).
 
       The `Class.TextChatMessage` parameter is the final result of any functions
-      bound to `Class.TextChatService.OnIncomingMessage` or
-      `Class.TextChannel.OnIncomingMessage`.
+      bound to `Class.TextChatService.OnIncomingMessage` or `Class.TextChannel.OnIncomingMessage`.
     code_samples: []
     parameters:
-      - name: incomingMessage
+      - name: textChatMessage
```
IgnisRBX, line 136:
```suggestion
      - name: incomingMessage
```

IgnisRBX (dismissed):

IgnisRBX (approved):

IgnisRBX:
I made a final round of slight edits, could you please review and let me know if it's good now? I can then merge and release. Thanks!

GFinkGH:
Looks good, although I had changed one thing earlier today which is the parameter name for `TextChannel.MessageReceived`. The TextChatService and TextChannel versions of this event share the same parameter class (`TextChatMessage`), but the docs page show them with a different name. [`TextChatService.MessageReceived`](https://create.roblox.com/docs/reference/engine/classes/TextChatService#MessageReceived) calls it `textChatMessage` while [`TextChannel.MessageReceived`](https://create.roblox.com/docs/en-us/reference/engine/classes/TextChannel#MessageReceived) calls it `incomingMessage`, yet for the `OnIncomingMessage` callback both instances call this parameter just 'message' — so maybe change that parameter name for both `MessageReceived` events to 'message' also. 

I am still curious about the copy-paste thing going on here. Since the TextChatService and TextChannel page share a lot of documentation, is it a foregone conclusion to copy and paste info between the two pages? Maybe a base class or something would work here, in the same way that `SpawnLocation` and `Part` share a lot of properties and methods through their abstract class `BasePart`. Or if that doesn't work, some central place that both of these articles pull from to fill in that information automatically without needing to copy and paste.

IgnisRBX:
Yep, I saw your request to change the name of `textChatMessage` to something like `incomingMessage`. Unfortunately the system is fairly strict and only allows docs to change summary, description, and deprecation_message fields... everything else will just be overwritten by the source.

As for the copy-paste thing for the two `MessageReceived` callbacks, I agree it's a bit redundant but that is how it was engineered and I'm not able to change that (without basically the entire API being rewritten internally lol). That does result in some copy-pasting which would ultimately be better as a shared snippet; I'll discuss this with engineering and see if there's a future solution.

Thanks again for your efforts on this fix; looking forward to any future contributions you may have on these classes and perhaps others as well. :)

IgnisRBX (approved):

PR #462, opened 2024-03-12
Clarify explosion behavior

BusyCityGuy:
## Changes
Explosion summary claimed that explosions are automatically `destroyed` after a short time. However, terminology is important here, as in practice explosions are only `removed`, so their parent is not locked and the connections to the explosion are not disconnected. By reading `destroyed`, readers may fail to understand this nuance, leading to them creating memory leaks due to not cleaning up their connections.

Related devforum topic: https://devforum.roblox.com/t/code-sample-for-explosionhit-has-a-memory-leak/2873076/6

## Checks


content/en-us/reference/engine/classes/Explosion.yaml
```diff
@@ -15,9 +15,7 @@ description: |
   `Class.Humanoid` characters not protected by a `Class.ForceField`.
   `Class.Constraint|Constraints` will not be broken by an explosion.
 
-  If an explosion is instanced while the game is running, it will destroy itself
-  shortly afterwards, so they do not need to be cleaned up using the
-  `Class.Debris` service.
+  If an explosion is parented while the game is running, it will automatically play the explosion effects and unparent itself when it finishes. Note that it is not destroyed, so connections do not get disconnected and the parent is not locked. If you hold open a connection containing a strong reference to the explosion, it will not be garbage collected.
```
IgnisRBX, line 18:
```suggestion
  If an explosion is parented while the experience is running, it will automatically play the explosion effects and unparent itself when it finishes. It is not destroyed in this case, so connections do not get disconnected and the parent is not locked. If you hold open a connection containing a strong reference to the explosion, it will not be garbage collected.
```

IgnisRBX: The last sentence, does that imply something even more specific to Explosion than typically holding a strong reference to a connection that would prevent garbage collection?

IgnisRBX: Also, in the first sentence, does it matter where it's parented to? If so, perhaps we should be more specific...

BusyCityGuy: Clarified the parent can be anywhere in the data model, and that garbage collection is not unique to this instance. I think it's still valuable to include a reminder here to give further meaning to why the distinction that it's not destroyed is important, given that this is a common pitfall for memory leaks.

FellowMeerkatRblx:
Hello there!

I'm coming from the linked topic, thank you for making the PR. I hope two small suggestions are not a nuisance.

- We could put it even more explicitly that explosions set off anywhere during runtime, but the effects happen only in workspace.
- The last two sentences apply to any instance that isn't destroyed. A reader should already know they have to lose the variable refering to the explosion, but I agree that some form of a reminder might serve well covering this particularly tricky case:

```lua
do local int = Instance.new("IntValue")
    int.Changed:Connect(function(newValue)
        int.Name = newValue
    end)
end
```
--> If the callback wasn't referring to the instance, this wouldn't leak memory.

Based on this we could write:

If an Explosion is parented anywhere in the data model while the experience is running, it sets off immediately and becomes unparented soon after the blast; as a descendant of `Class.Workspace`, it plays the explosion effects. Since an Explosion is not destroyed automatically, make sure to disconnect any connections that may prevent it from being released from memory and, consequently, keep any other connections to it still active, or to destroy the Explosion.

BusyCityGuy:
> If an Explosion is parented anywhere in the data model while the experience is running, it sets off immediately and becomes unparented soon after the blast; as a descendant of `Class.Workspace`, it plays the explosion effects. Since an Explosion is not destroyed automatically, make sure to disconnect any connections that may prevent it from being released from memory and, consequently, keep any other connections to it still active, or to destroy the Explosion.

The flow of this description is a bit hard to follow. If we want to include the detail of no effects outside workspace and be even more clear about avoiding the common pitfall, I would propose something like this:

>  If an explosion is parented to `Class.Workspace` while the experience is running, it automatically plays the explosion effects and unparents itself once the effects finish. If parented to the data model outside of Workspace, no effect is played but it still unparents itself after a few seconds. In either case, the explosion is not destroyed with `Class.Instance:Destroy()`. As with all instances, keeping a strong reference the explosion will prevent it from being garbage collected. A common pitfall leading to a memory leak is referencing the explosion inside a `.Hit` connection without later disconnecting the connection or destroying the explosion.

FellowMeerkatRblx:
> The flow of this description is a bit hard to follow.

Alright, fair point.

I just discovered that Explosion's lifetime depends on its BlastRadius regardless of the Parent or TimeScale. Blast radius 1 stud --> ~2 seconds; radius (max) 100 studs --> ~3 seconds. If TimeScale is very low, the visuals are cut off. I don't know whether to classify this as a bug, but it seem to be delicate legacy behaviour just like unparenting.

Explosions also explode when parented to any replicable descendant of Workspace, including Terrain, and even very odd places like constraints or sound and sound effect instances.

As for the other part of the paragraph, your wording seems quite fine. We'll see how the maintainers decide since it applies to strong references in general, though in this case we're dealing with legacy code.

---

Small proposition for the first part of the paragraph:

If an Explosion is parented anywhere in the data model while the experience is running, it immediately sets off. Within a few seconds, it becomes unparented. However, an Explosion must be a descendant of `Class.Workspace` for the explosion effects to play and have impact.

FellowMeerkatRblx:
> The flow of this description is a bit hard to follow.

Alright, fair point.

I just discovered that Explosion's lifetime depends on its BlastRadius regardless of the Parent or TimeScale. Blast radius 1 stud --> ~2 seconds; radius (max) 100 studs --> ~3 seconds. If TimeScale is very low, the visuals are cut off after the period.

Explosions also explode when parented to any replicable descendant of Workspace, including Terrain, and even very odd places like constraints or sound and sound effect instances.

As for the other part of the paragraph, your wording seems quite fine. We'll see how the maintainers decide since it applies to strong references in general, though in this case we're dealing with legacy code.

---

Small proposition for the first part of the paragraph:

If an Explosion is parented anywhere in the data model while the experience is running, it immediately sets off. Within a few seconds, it becomes unparented. However, an Explosion must be a descendant of `Class.Workspace` for the visuals to play and explosion effects to have an impact.

BusyCityGuy:
> I just discovered that Explosion's lifetime depends on its BlastRadius regardless of the Parent or TimeScale. Blast radius 1 stud --> ~2 seconds; radius (max) 100 studs --> ~3 seconds. If TimeScale is very low, the visuals are cut off after the period. So the phrasing depends on whether the particle effect counts among explosion effects or not.

I tried correlating BlastRadius with lifetime and got some interesting results.
image

There's a positive correlation value of .4724293617473616, so it's not completely correlated but it _is_ correlated. Great find.

I took this a step further and I think I figured out how this works.

image

Below BlastRadius 1, it seems to just use 3. Above BlastRadius 1, it seems to fit the curve 3 - 1 / BlastRadius. The noisy results that are above the line can be attributed to the scheduler not always resuming the thread immediately.

So this does not account for the TimeScale value at _all_, my guess is it was added after the original timer code for the explosion was written.

BusyCityGuy:
Maybe @IgnisRBX would be willing to come up with best wording taking all of the above into account

IgnisRBX:
Do we consider it important enough to document the timer inconsistency? Are developers encountering this and having issues understanding how to resolve it?

content/en-us/reference/engine/classes/Explosion.yaml
```diff
@@ -228,7 +231,7 @@ properties:
     summary: |
       Value between 0-1 that controls the speed of the particle effect.
```
IgnisRBX, line 232:
```suggestion
      Value between 0 and 1 that controls the speed of the particle effect.
```

content/en-us/reference/engine/classes/Explosion.yaml
```diff
@@ -228,7 +231,7 @@ properties:
...
     description: |
-      A value between 0-1 than controls the speed of the particle effect. At 1
+      A value between 0-1 that controls the speed of the particle effect. At 1
```
IgnisRBX, line 234:
```suggestion
      A value between 0 and 1 that controls the speed of the particle effect. At 1
```

BusyCityGuy:
> Do we consider it important enough to document the timer inconsistency? Are developers encountering this and having issues understanding how to resolve it?

I do not think timer inconsistency needs to be documented, since it is a bug (I reported it [here](https://devforum.roblox.com/t/explosions-cleanup-timer-is-too-primitive-to-comprehensively-wait-for-vfx-time/2876299))

The important part to document is how it only gets parented to `nil`, not destroyed. 
And the part that I wanted you to figure out best wording for is the wording around when plays effects & does the deparenting depending on the container it's in. Comments above have some proposals

content/en-us/reference/engine/classes/Explosion.yaml
```diff
@@ -15,30 +15,33 @@ description: |
...
 
-  If an explosion is instanced while the game is running, it will destroy itself
-  shortly afterwards, so they do not need to be cleaned up using the
-  `Class.Debris` service.
+  If an explosion is parented to the data model while the experience is
+  running, it will automatically play the explosion effects and unparent
+  itself when it finishes. It is not destroyed with `Class.Instance:Destroy()`
+  in this case, so connections do not get disconnected and the parent is not
+  locked. As with all instances, keeping a strong reference the explosion
+  will prevent it from being garbage collected.
```
IgnisRBX, line 23:
```suggestion
  If an `Class.Explosion` is parented anywhere in the data model while the
  experience is running, it immediately sets off and, within a few seconds,
  it becomes unparented. It is not destroyed with `Class.Instance:Destroy()`
  in this case, so connections do not get disconnected and the parent is not
  locked. As with all instances, keeping a strong reference an
  `Class.Explosion` will prevent it from being garbage collected.
  
  Note that an `Class.Explosion` must be a descendant of `Class.Workspace`
  for the explosion visuals to play and the physical/damaging effects to have
  an impact.
```

IgnisRBX (approved):

PR #463, opened 2024-03-14
More accurate debris destroy behavior

BusyCityGuy:
## Changes
The current documentation has some issues:
* It uses `wait` and `delay` instead of `task.wait` and `task.delay`
* It duplicates the long summary in the `AddItem` method
* It does not document what the `MaxItems` value is, and does not mention `MaxItems` behavior in the summary or `AddItem` description despite `MaxItems` having an important role in how `Debris` works
* It uses the word `remove` instead of `destroy` in several places, but it should use `destroy` because `remove` is a different concept on Roblox
* It uses a projectile as an example of a good use for Debris, but it is not, because Debris does not gaurantee the exact lifetime of the object
* It states Debris is the recommended way to clean up instances, but it should not be recommended because [it will likely be deprecated](https://devforum.roblox.com/t/debris-maxitems-still-in-effect-despite-being-deprecated/2612863/3)
* It makes a point of differentiating `Debris` from `Destroy` by saying `Destroy` will error if called on an already-destroyed instance, which is incorrect
* It does not mention one important valuable differentiation from `Destroy`, that `Debris` runs outside the context of the script, so it guarantees destruction of the object whereas scheduling a `Destroy` call depends on the script continuing to exist and be enabled

## Checks


content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,73 @@ type: class
 category: Utility
 memory_category: Instances
 summary: |
-  The Debris service allows the developer to schedule the removal of the object
-  without yielding any code, through the usage of the `Class.Debris:AddItem()`
+  The Debris service allows the developer to schedule guaranteed destruction of
+  the object without yielding, through the usage of the `Class.Debris:AddItem()`
   method.
```
IgnisRBX, line 8:
```suggestion
  Allows scheduling of guaranteed destruction of an object without yielding.
```

IgnisRBX: IMHO summaries can be a bit shorter and not mention other methods within the class.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,73 @@ type: class
...
 description: |
-  The Debris service allows the developer to schedule the removal of the object
-  without yielding any code, through the usage of the `Class.Debris:AddItem()`
+  The Debris service allows the developer to schedule guaranteed destruction of
+  the object without yielding, through the usage of the `Class.Debris:AddItem()`
```
IgnisRBX, line 11:
```suggestion
  The **Debris** service allows scheduling of guaranteed destruction of an
  object without yielding, through the usage of `Class.Debris:AddItem()`.
```

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,73 @@ type: class
...
   method.
 
-  After the lifetime argument has elapsed (in seconds) the object is removed in
-  the same manner as `Class.Instance:Destroy()`.
+  After the lifetime argument has elapsed (in seconds) the object is destroyed
+  in the same manner as `Class.Instance:Destroy()`.
 
   As Debris is a service it must be created using the
   `Class.ServiceProvider:GetService()` method.
```
IgnisRBX, line 18:
```suggestion
```

IgnisRBX: Can we omit this? I don't recall seeing other services making this specific usage note.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
...
 
   ```lua
-  delay(3, function()
-  	if projectile and projectile.Parent then
-  		projectile:Destroy()
-  	end
+  task.delay(3, function()
+  	brick:Destroy()
   end)
   ```
```
IgnisRBX, line 43: I took out the second suggestion of doing this "all on one line" because it seems that one "bad" example is enough. However, if you think it's useful to show both ways of calling task.delay(), feel free to add it back in.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
...
   method.
 
-  After the lifetime argument has elapsed (in seconds) the object is removed in
-  the same manner as `Class.Instance:Destroy()`.
-
   As Debris is a service it must be created using the
   `Class.ServiceProvider:GetService()` method.
 
-  **Why use Debris?**
+  #### Advantages
```
IgnisRBX, line 17: Mostly because I dislike the tone of "asking questions" in technical docs. :)

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
...
 
   ```lua
-  Debris:AddItem(projectile, 3)
+  Debris:AddItem(brick, 3)
   ```
 
-  Debris does not yield the current thread, does not require a new thread and
-  will not error if the object is already destroyed. For this reason it is the
-  recommended method for cleaning up objects with a fixed lifetime.
+  Note that `Class.Debris` has a hardcoded maximum of 1,000 objects, so if
```
IgnisRBX, line 58: Are we fairly certain the number is still a hardcoded 1000, and not some internal flexible value based on occupied memory, device memory, or some other factors?

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -99,74 +97,32 @@ properties:
 methods:
   - name: Debris:AddItem
     summary: |
-      Allows the developer to schedule the removal of the object without
-      yielding any code.
+      Registers a given `Class.Instance` for removal within the specified
```
IgnisRBX, line 100: The previous summary sounded more like a summary for the Debris service in general, and not specifically for AddItem().

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
...
+  the object without yielding, through the usage of the `Class.Debris:AddItem()`
   method.
 description: |
-  The Debris service allows the developer to schedule the removal of the object
-  without yielding any code, through the usage of the `Class.Debris:AddItem()`
+  The **Debris** service allows scheduling of guaranteed destruction of an
+  object without yielding, through the usage of the `Class.Debris:AddItem()`
   method.
 
-  After the lifetime argument has elapsed (in seconds) the object is removed in
```
IgnisRBX, line 14: I took this out here, and added it into the AddItem() description, since it describes a behavior specific to that method.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Yes, this is just hardcoded 1,000 and does not fluctuate with any factors. Attempting to set it to a different number will error.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Yeah it was weird, the entire service summary was duplicated in the AddItem summary.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Makes sense!

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: I agree, I suppose it was included before just because there is only one useful api member, but it doesn't seem necessary to me to call out here.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Just left it in from before, but yes I agree it doesn't need to be here if it's not on other service pages.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Just kept the header from before, but yeah thanks for keeping the docs consistent and neutral

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: Removing the first `of`, it reads better I think
> Allows scheduling guaranteed destruction of an object without yielding.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy: I think it is useful to keep, because for some people, they might not know that there is a 1-line solution, which could be a downside to the reader compared to `task.delay`. Including both makes it clear what's happening with the multi-line solution, and clear that there is still a 1-line option if you want to use debris.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
...
+  introduces a second drawback in that the scheduled callback will never run if
+  the script is disabled or destroyed before the callback runs.
 
-  This is where Debris comes in, and the following code addresses all of the
-  above issues.
+  This is where `Class.Debris` has a specific advantage, as it does not yield
+  the current thread and runs outside the context of the script, guaranteeing
+  the instance is eventually destroyed even if the script is disabled or
+  destroyed. The following code addresses both issues cited for
+  `Library.task.wait()` and `Library.task.delay()`:
```
Nightriff, line 52:
```suggestion
This is where `Class.Debris` has a specific advantage, as it does not yield the current thread and it runs outside the context of the script, guaranteeing the instance is eventually destroyed even if the script is disabled or destroyed. The following code does not yield and gaurantees the instance will be destroyed:
```

`task.wait()` has both issues (yields + depends on calling thread to live)
`task.delay()` has only one issue (does not yield, but depends on calling thread to live)
`Debris:AddItem` has neither (does not yield and does not depend on calling thread to live)

They're not necessarily really "issues", just differences in behavior. Perhaps you do want it to yield, or perhaps you do want it to not be destroyed if the script is disabled.

content/en-us/reference/engine/classes/Debris.yaml
BusyCityGuy:
```suggestion

  Or in one line:
  
  ```lua
  task.delay(3, brick.Destroy, brick)
  ```

```

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,59 @@ type: class
...
+
+  While this now avoids yielding, it has a drawback in that the scheduled callback will never run if the script is disabled or destroyed before the callback runs.
 
-  This is where Debris comes in, and the following code addresses all of the
-  above issues.
+  This is where `Class.Debris` has a specific advantage, as it does not yield
+  the current thread and runs outside the context of the script, guaranteeing
+  the instance is eventually destroyed even if the script is disabled or
+  destroyed. The following code addresses both issues cited for
+  `Library.task.wait()` and `Library.task.delay()`:
```
BusyCityGuy, line 51:
```suggestion
This is where `Class.Debris` has a specific advantage, as it does not yield the current thread and it runs outside the context of the script, guaranteeing the instance is eventually destroyed even if the script is disabled or destroyed. The following code does not yield and gaurantees the instance will be destroyed:
```

`task.wait()` has both issues (yields + depends on calling thread to live)
`task.delay()` has only one issue (does not yield, but depends on calling thread to live)
`Debris:AddItem` has neither (does not yield and does not depend on calling thread to live)

They're not necessarily really "issues", just differences in behavior. Perhaps you do want it to yield, or perhaps you do want it to not be destroyed if the script is disabled.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,59 @@ type: class
...
```
BusyCityGuy, line 50:
```suggestion
The following code does not yield and guarantees the instance will be destroyed:
```

`task.wait()` has both issues (yields + depends on calling thread to live)
`task.delay()` has only one issue (does not yield, but depends on calling thread to live)
`Debris:AddItem` has neither (does not yield and does not depend on calling thread to live)

They're not necessarily really "issues", just differences in behavior. Perhaps you do want it to yield, or perhaps you do want it to not be destroyed if the script is disabled.

BusyCityGuy, line 50:
```suggestion
  destroyed. The following code does not yield and guarantees the instance will
  be destroyed:
```

`task.wait()` has both issues (yields + depends on calling thread to live)
`task.delay()` has only one issue (does not yield, but depends on calling thread to live)
`Debris:AddItem` has neither (does not yield and does not depend on calling thread to live)

They're not necessarily really "issues", just differences in behavior. Perhaps you do want it to yield, or perhaps you do want it to not be destroyed if the script is disabled.

content/en-us/reference/engine/classes/Debris.yaml
```diff
@@ -3,64 +3,61 @@ type: class
 category: Utility
 memory_category: Instances
 summary: |
-  The Debris service allows the developer to schedule the removal of the object
-  without yielding any code, through the usage of the `Class.Debris:AddItem()`
-  method.
+    Allows scheduling guaranteed destruction of an object without yielding.
```
IgnisRBX, line 6:
```suggestion
    Allows scheduling the guaranteed destruction of an object without yielding.  
```

IgnisRBX (approved):

PR #470, opened 2024-03-19
Document order of PlayerAdded and ProcessReceipt

nezuo:
## Changes

- Added a note about the order of when `PlayerAdded` is fired for a player and for when `ProcessReceipt` is called for that player.

I don't know if this order *is* guaranteed but I couldn't find otherwise when testing. 

## Confusing Docs
One of the notes under `Garauntees` states:
> The player needs to be in the server for the callback to be invoked...

However, the docs check to see if the player is still in the game:
https://github.com/Roblox/creator-docs/blob/c6e1aca2cec792be93d9062ba5f1ed1ccadc3914/content/en-us/production/monetization/developer-products.md?plain=1#L122-L127
The "Implementing Player Data and Purchasing Systems" example also has this check:
```lua
function ReceiptProcessor._processReceiptAsync(
	playerId: number,
	productId: number,
	purchaseId: string
): Enum.ProductPurchaseDecision
	-- We do not want to save player data if the player is not currently in the server
	local player = Players:GetPlayerByUserId(playerId)
	if not player then
		return Enum.ProductPurchaseDecision.NotProcessedYet
	end
```

I was wondering if this "guarantee" is not true or if the two examples are accounting for an edge case that doesn't exist.

If it is guaranteed, I would like to see the note mention that `Players:GetPlayerByUserId(receiptInfo.PlayerId)` is guaranteed (assuming the user hasn't yielded) to return the `Player`. I can make that change if this is the case.

## Checks


content/en-us/reference/engine/classes/MarketplaceService.yaml
```diff
@@ -1198,6 +1198,9 @@ callbacks:
         does not need to be in the server for the result of the callback to be
         recorded on the backend.
 
+      - The callback is called for a player after the PlayerAdded event has
```
IgnisRBX, line 1201:
```suggestion
      - The callback is called for a player after the `Class.Players.PlayerAdded` event has
```

IgnisRBX (approved):

IgnisRBX (changes requested):
Awaiting further review by @othomson-roblox

IgnisRBX:
@othomson-roblox quick ping on this.

PR #505, opened 2024-04-13
"Provide Keywords" good example should be less confusing

VoiditeDev:
## Changes

This is a small change as it only changes one word. but in this good example, I got confused by "become a tycoon"
[image](https://create.roblox.com/docs/production/publishing/publishing-experiences-and-places#experience-description)
It would've been less confusing to change "become" to "start". also ignore any commits that isn't this:

image

## Checks


aetter:
Hey @VoiditeDev, thanks for the PR, but I don't agree that the original wording is confusing. A tycoon is a businessperson, not a business, so you would indeed become a tycoon rather than starting one.

VoiditeDev:
@aetter I will undo the request I have made

VoiditeDev:
oh, also I would recommend changing the word from "tycoon" to "businessman" to make it less confusing

PR #514, opened 2024-04-23
Improvements to Instance.Parent documentation

EmeraldSlash:
## Changes

Added a common piece of undocumented knowledge: that changing properties while an instance is parented can be expensive and trigger more work.

e.g. https://devforum.roblox.com/t/psa-dont-use-instancenew-with-parent-argument/30296

But also based on my observations from using the engine and from looking at the microprofiler and seeing things like GUI layouting work being performed after reading/changing a property.

Also improved the wording of the Object Replication section so that it is clearer, as I couldn't understand what it is was saying the first time I read it.

## Checks


EmeraldSlash:
> Hey @EmeraldSlash, it seems like this is covered pretty well in the Object Replication section right below the new paragraph. Is there some additional nuance here that I'm missing?

That section only refers to replication. But there are other cases which have nothing to do with replication where parenting order is important and where changing properties can incur additional work. So that's why I think it's a good idea to document that behaviour more generally.

I agree as it's written now there's a bit of duplication.. I wanted to keep the changes minimal and not rewrite the whole section or remove emphasis on the replication part which is still important.

content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -191,9 +191,16 @@ properties:
       Instance.new("NumberValue", workspace)
       ```
 
+      Whena a change is made to certain properties while an instance is
```
aetter, line 194:
```suggestion
      When a change is made to certain properties while an instance is
```

content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -191,9 +191,16 @@ properties:
...
+      parented in the `Class.DataModel` the engine may need to perform extra
```
aetter, line 195:
```suggestion
      parented in the `Class.DataModel`, the engine may need to perform extra
```

content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -191,9 +191,16 @@ properties:
...
+      work internally (for things like replication, rendering, and GUI
+      layouting). Therefore where possible it is a good idea to change an
```
aetter, line 197:
```suggestion
      layout). Whenever possible, change an
```

aetter (approved):

aetter (approved):

aetter:
Sounds good, thanks. I might give it a pass later to try and deduplicate, but that's no reason not to take the improvement now. Thanks again!

PR #573, opened 2024-05-31
Fix typo

birds3345:
that -> that's


IgnisRBX:
Hi there! We're going to fix this in a separate pull request, so I'm going to close this one... but thank you for flagging this! Any other bits like this that you notice, please open another pull request and we'll take care of it.

PR #619, opened 2024-06-11
Correct module return information

Someon1e:
## Changes

Removed incorrect statement that modules can not return nil.
image

## Checks


content/en-us/scripting/scripts.md
```diff
@@ -28,6 +28,7 @@ load by calling the `Global.RobloxGlobals.require()` function. Module scripts mu
 value and run once and only once
 per Lua environment. As a result, subsequent calls to `Global.RobloxGlobals.require()` return a
 cached value.
+You can execute arbitrary code in a `Class.ModuleScript`, but you only need to return what you need in other scripts.
```
IgnisRBX, line 31:
```suggestion
cached value. You can execute arbitrary code in a `Class.ModuleScript`, but you only need to return what you need in other scripts.
```

IgnisRBX (approved):

PR #643, opened 2024-06-20
Clarified `Instance.Changed`.

Windows81:
## Changes

Clarified `Instance.Changed`.

## Checks


content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -1646,9 +1646,16 @@ events:
...
+      This event fires immediately after an instance property is changed
+      and works with most use cases (see Limitations below). The new value
+      of a changed property is *not* passed in as a parameter.
+      Therefore, it must be accessed by using `object[property]`.  For example:
+      
+      ```
+      object.Changed:Connect(function(property)
+      	print("The new property's value is", object[property])
+      end)
+      ```
```
IgnisRBX, line 1658:
```suggestion
      This event fires immediately after an instance property is changed
      and it works with most use cases (see limitations below). The new value
      of a changed property is **not** passed in as a parameter, so it must
      be accessed by using `object[property]`. For example:
      
      ```
      object.Changed:Connect(function(property)
      	print("The new property's value is", object[property])
      end)
      ```
```

content/en-us/reference/engine/classes/Instance.yaml
```diff
@@ -1646,9 +1646,16 @@ events:
     summary: |
       Fired immediately after a property of the instance changes.
```
IgnisRBX, line 1647:
```suggestion
      Fired immediately after a property of the instance changes, with some limitations.
```

IgnisRBX (approved):

PR #650, opened 2024-06-22
Update UserInputService.yaml

lack-of-gravity-jack:
## Changes

"undetermined instead of a vector2" implies it will return nil if the mouse goes offscreen
it is indeed undetermined but the "instead' is misleading, it will always return a vector2 from testing.

## Checks

- [x ] This contribution was created in whole or in part by me, and I have the right to submit it under the terms of this repository's open source licenses.
- [x ] I understand and agree that this contribution and a record of it are public, maintained indefinitely, and may be redistributed under the terms of this repository's open source licenses.
- [x ] To the best of my knowledge, all proposed changes are accurate.


content/en-us/reference/engine/classes/UserInputService.yaml
```diff
@@ -1113,8 +1113,7 @@ methods:
       corner. This does not account for the `Class.GuiObject|GUI` inset.
 
       If the location of the mouse pointer is offscreen or the players device
```
IgnisRBX, line 1115:
```suggestion
      If the location of the mouse pointer is offscreen or the player's device
```

content/en-us/reference/engine/classes/UserInputService.yaml
```diff
@@ -1113,8 +1113,7 @@ methods:
...
-      does not have a mouse, the value returned will be undetermined instead of
-      Vector2.
+      does not have a mouse, the value returned will be undetermined.
```
IgnisRBX, line 1116:
```suggestion
      does not have a mouse, the returned value will be undetermined.
```

IgnisRBX (approved):

PR #683, opened 2024-07-06
Change wording regarding require(id) in ModuleScript.yaml

metatablecat:
## Changes

Private modules have not existed since 2019, although the logic is still there, this was written with the intent to share the code to other users.

## Checks


content/en-us/reference/engine/classes/ModuleScript.yaml
```diff
@@ -37,10 +37,11 @@ description: |
...
 
-  If a ModuleScript object has its Name property set to 'MainModule' and is
-  uploaded to Roblox as a model to your account, Scripts can use `require` with
-  the uploaded model's AssetId instead. This allows you to create private
-  modules on your Roblox account!
+  If a ModuleScript is uploaded to Roblox, and the root module has the name set
+  to `MainModule`, it can be uploaded as a model, and required using 
+  `Global.RobloxGLobals.require()` with the model's asset ID, it can be loaded
+  into your experience. This logic only works on the server, and will error on
+  the client. If other users want to use the module, **it must be public**.
```
IgnisRBX, line 44:
```suggestion
  If a `Class.ModuleScript` is uploaded to Roblox and the root module has the name
  set to `MainModule`, it can be uploaded as a model and required using 
  `Global.RobloxGLobals.require()` with the model's asset ID. Then it can be loaded
  into your experience, although this logic only works on the server and will error on
  the client. If other users want to use the module, **it must be public**.
```

IgnisRBX (approved):

PR #695, opened 2024-07-09
Remove occurences of Class.CoreScript reference

Etheroit:
## Changes

CoreScript is not a class but a concept of specially privileged script, there is nothing to be referenced (especially this wouldn't be a non-existing `CoreScript` class). So far there on the creator hub documentation there is no explanation what CoreScripts really are and there doesn't seem to be any place for that at the moment.

## Checks


Etheroit:
Alright - actually I'm wrong. CoreScript class exists - it's just not documented on Creator Hub

PR #720, opened 2024-07-17
ypcall-addition-and-deprecation-messages

daskairo:
## Changes

- Minor format changes referencing Classes and Functions.
- Added Deprecation messages to `collectgarbage()` to direct readers to `gcinfo()`.
- Added missing `ypcall()` function to page with Deprecation Tag and Message, with it directing readers to `pcall()`.

> I am unsure if `ypcall()` falls under "Roblox Globals" or "Lua Globals", but as `pcall()` and `xpcall()` exist under "Lua Globals", It would probably be better there. Though, that I feel that may make the documentation inaccurate n some way.

Luau Sandboxing: https://luau-lang.org/sandbox#library
DevForum Description: https://devforum.roblox.com/t/1417471
Roblox Forum Archive: https://archive.froast.io/forum/73503253
2012 Roblox Wiki Archive: https://web.archive.org/web/20121022200149/http://wiki.roblox.com/index.php/Function_Dump/Roblox_Specific_Functions#ypcall
ScriptingHelpers Desription: https://scriptinghelpers.org/questions/39638

## Checks


content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -725,3 +728,28 @@ functions:
         summary: ''
     tags:
     code_samples:
+  - name: ypcall
+    summary: |
+      Similar to `pcall()` expect it yields.
+    description: |
+      This function is similar to `pcall()`, except that you can yield function `f` to be called in protected mode.
```
vegorov-rbx, line 735: This is incorrect, you can yield in `pcall`.

content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -725,3 +728,28 @@ functions:
         summary: ''
     tags:
     code_samples:
+  - name: ypcall
+    summary: |
+      Similar to `pcall()` expect it yields.
```
vegorov-rbx, line 733: This is incorrect, `ypcall` doesn't yield by itself. In fact, it is absolutely identical to `pcall`.

daskairo:
Dug through the Internet Archive and found a reference to `ypcall()` within the 2012 Roblox Wiki snapshot.

Added that description with slight changes in wording and format.

content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -692,7 +691,7 @@ functions:
     summary: |
       Similar to `pcall()` except it uses a custom error handler.
     description: |
-      This function is similar to pcall, except that you can set a new error
+      This function is similar to `pcall()`, except that you can set a new error
```
IgnisRBX, line 694:
```suggestion
      This function is similar to `Global.LuaGlobals.pcall()`, except that you can set a new error
```

content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -692,7 +691,7 @@ functions:
     summary: |
       Similar to `pcall()` except it uses a custom error handler.
```
IgnisRBX, line 692:
```suggestion
      Similar to `Global.LuaGlobals.pcall()` except it uses a custom error handler.
```

vegorov-rbx:
> Dug through the Internet Archive and found a reference to `ypcall()` within the 2012 Roblox Wiki snapshot.

This doesn't reflect how it works on Roblox today.

daskairo:
> This doesn't reflect how it works on Roblox today.

This doesnt matter, and you could apply this any other deprecated item from years ago.

Its better to keep the information of how it worked at the time, what it did in return for readers to understand it.
Because it no longer retains the same functionality and is now identical to `pcall()`, its best that we redirect the reader to a non deprecated alternative.

Its only fair that people get to know its original use, because for a long time `ypcall()` was missing any documentation.

vegorov-rbx (changes requested):
There is no difference between pcall and ypcall on Roblox and documentation should reflect reality to not confuse anybody.

vegorov-rbx:
I disagree and PR will be blocked until the issue is fixed.

dphblox:
Hmm. I definitely agree that we should be documenting current behaviour rather than historical behaviour. That said they do have a point about people being confused by historical usage. Perhaps the way to thread that needle is to document it as something like

> Identical to `pcall()` and should not be used for new work. A legacy function to work around an old task scheduling limitation of `pcall`.

That's as far as I'd ever go with it. There's no point in documenting what something did in 2012 if it isn't notable now. We should make sure that developers can modernise legacy codebases, so including a little context is fine for the purpose of allowing them to do that, but a full blown description of things that *don't* happen isn't ultimately useful for most people whom the docs are catered towards - at least, that's my personal take.

daskairo:
> Hmm. I definitely agree that we should be documenting current behaviour rather than historical behaviour. That said they do have a point about people being confused by historical usage. Perhaps the way to thread that needle is to document it as something like
> 
> > Identical to `pcall()` and should not be used for new work. A legacy function to work around an old task scheduling limitation of `pcall`.
> 
> That's as far as I'd ever go with it. There's no point in documenting what something did in 2012 if it isn't notable now - creator docs isn't meant to be an archival record but instead a reference for everyday development. We should make sure that developers can modernise legacy codebases, so including a little context is fine for the purpose of allowing them to do that, but a full blown description of things that *don't* happen isn't ultimately useful for most people whom the docs are catered towards - at least, that's my personal take.

That's pretty fair, I just think that people should know what it originally did and why it was replaced/deprecated. Maybe it no longer works, maybe it was retired, maybe it was superseded. Typically that kind of message.

In my opinion, it would be better to document deprecated items, mostly to explain:

- what they did
- why they are deprecated
- what's the alternative?

Generally in case people stumble upon something, they can read into and find the proper replacement.
The only time I would agree that "Hey this should be here" is If it was outright removed, for which year, there shouldn't be any documentation on it, as it no longer exists.

That's just my take, people may not agree but that's fine. To me, documentation is a place for people to read into an assortment of things (deprecated or not) to learn, and improve.

dphblox:
I appreciate that there's a lot to learn from the history of how things have worked on platform, and as a veteran myself, it would be wonderful for people to pursue their interest in learning about the platform's history. 

That said, If we include too much trivia in the documentation, then we run the risk of confusing new users or presenting topics as overcomplicated, which can intimidate away new learners or muddy up the concepts we try and teach them. So for those reasons, while it's fantastic that you've gone to this effort to collect this information together, we likely won't be able to accept all of it.

If I may suggest though, there's plenty of wikis out there which are much more outwardly dedicated to these archival-style topics. This would be fantastic to compile over there :)

We would love to respect your work (and truthfully, we're very glad that you're here!) - but we're just trying to balance it up against the vast responsibility that our docs serve to our large community of creators.

content/en-us/reference/engine/globals/RobloxGlobals.yaml
```diff
@@ -423,3 +423,30 @@ functions:
         summary: ''
     tags:
     code_samples:
+  - name: ypcall
+    summary: |
+      Similar to `pcall()` expect it can yield.
+    description: |
+      This function is similar to `pcall()`. However, the function passed as an argument can yield without causing error (that is, the "attempt to yield across metamethod/C-call boundary" error).
+      
+      This allows functions such as `wait()` to be used. In addition, should the function passed error, the error is printed into the output although the code that called `ypcall()` remains unaffected.
```
IgnisRBX, line 432:
```suggestion
     A legacy function to work around an old task scheduling limitation of `Global.LuaGlobals.pcall()`;
     should not be used for new work (use `Global.LuaGlobals.pcall()` instead).
```

content/en-us/reference/engine/globals/RobloxGlobals.yaml
```diff
@@ -423,3 +423,30 @@ functions:
...
+        summary: ''
+    returns:
+      - type: bool
+        summary: ''
+      - type: Variant
+        summary: ''
+    tags:
+      - Deprecated
+    deprecation_message: |
+      This function is now completely identical to `pcall()` which should be used instead.
```
IgnisRBX, line 451:
```suggestion
```

content/en-us/reference/engine/globals/RobloxGlobals.yaml
```diff
@@ -423,3 +423,30 @@ functions:
         summary: ''
     tags:
     code_samples:
+  - name: ypcall
+    summary: |
+      Similar to `pcall()` expect it can yield.
```
IgnisRBX, line 428:
```suggestion
      Legacy function to work around an old task scheduling limitation of `Global.LuaGlobals.pcall()`.
```

IgnisRBX (approved):

PR #726, opened 2024-07-21
Fix variadic `CFrame` methods

Ketasaja:
## Changes

These methods are not fully documented as being variadic.

I don't know how to correctly document a variadic return when all returned values have known type, and didn't find examples, so I went with Luau's type syntax `...CFrame`. Please tell me if this should instead be `Variant` or something else.

## Checks


content/en-us/reference/engine/datatypes/CFrame.yaml
```diff
@@ -743,120 +743,108 @@ methods:
     deprecation_message: ''
   - name: CFrame:ToWorldSpace
     summary: |
-      Returns one or more `Datatype.CFrame` objects transformed from object to
-      world space.
+      Recives one or more `Datatype.CFrame` objects and returns them transformed from object to world space.
```
IgnisRBX, line 746:
```suggestion
      Receives one or more `Datatype.CFrame` objects and returns them transformed from object to world space.
```

content/en-us/reference/engine/datatypes/CFrame.yaml
```diff
@@ -743,120 +743,108 @@ methods:
...
     description: |
-      Returns one or more `Datatype.CFrame` objects transformed from object to
-      world space. Equivalent to:
+      Receives one or more `Datatype.CFrame` objects and returns them transformed from object to world space. Equivalent to:
 
       `CFrame * cf`
     parameters:
-      - name: cf
+      - name: ...
         type: CFrame
```
IgnisRBX, line 753:
```suggestion
        type: Tuple
```

content/en-us/reference/engine/datatypes/CFrame.yaml
```diff
@@ -743,120 +743,108 @@ methods:
...
         default:
         summary: ''
     returns:
-      - type: CFrame
+      - type: ...CFrame
```
IgnisRBX, line 757:
```suggestion
      - type: Tuple<CFrame>
```

content/en-us/reference/engine/datatypes/CFrame.yaml
IgnisRBX: I think what you mean here is that it can take a tuple of CFrames, yes? And that it also returns a tuple of CFrames?

IgnisRBX: In this case, we wouldn't want to use `<Variant>` since neither the input nor output can be a tuple of mixed types (all must be CFrames and all coming back are CFrames).

aetter: This is super cool syntax -- wasn't aware we supported it.

IgnisRBX: Here's an example of where a tuple of numbers is cited:
https://create.roblox.com/docs/reference/engine/libraries/utf8#codepoint

IgnisRBX (approved):

PR #738, opened 2024-07-30
Update StudioService.yaml (code example ActiveScript)

KevinCodeCraft:
Previously, newActiveScript was used as a parameter, but GetPropertyChangedSignal does not accept parameters. As a result, newActiveScript was always nil, causing the script to fail. I have made newActiveScript a variable and I also added an existence check for ActiveScript.

### DIFF
--- a/content/en-us/reference/engine/classes/StudioService.yaml
+++ b/content/en-us/reference/engine/classes/StudioService.yaml
@@ -44,8 +44,10 @@ properties:
       ```lua
       local startTime = os.time()
       local activeScript
-      local function onActiveScriptChanged(newActiveScript)
-      	if newActiveScript ~= activeScript then
+      local function onActiveScriptChanged()
+        local newActiveScript = game:GetService("StudioService").ActiveScript
+      
+      	if newActiveScript ~= activeScript and activeScript then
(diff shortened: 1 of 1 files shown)


IgnisRBX:
I put top line reference to StudioService instead of getting the service twice later in the code.

IgnisRBX (approved):

content/en-us/reference/engine/classes/StudioService.yaml
```diff
@@ -42,17 +42,22 @@ properties:
...
+      local StudioService = game:GetService("StudioService")
+      
       local startTime = os.time()
       local activeScript
-      local function onActiveScriptChanged(newActiveScript)
-      	if newActiveScript ~= activeScript then
+      local function onActiveScriptChanged()
+        local newActiveScript = StudioService.ActiveScript
+      
+      	if newActiveScript ~= activeScript and activeScript then
```
IgnisRBX, line 52:
```suggestion
      	if activeScript and newActiveScript ~= activeScript then
```

IgnisRBX (approved):

KevinCodeCraft:
I understand, I did not touch it because it was like that so I assumed that it was preferred that way :)

PR #747, opened 2024-08-03
Fix `os.clock` description

Ketasaja:
"time used by Luau" is not correct.

https://github.com/luau-lang/luau/blob/ce8495a69e7a4e774a5402f99e1fc282a92ced91/VM/src/lperf.cpp#L57

### DIFF
--- a/content/en-us/reference/engine/libraries/os.yaml
+++ b/content/en-us/reference/engine/libraries/os.yaml
@@ -300,10 +300,10 @@ functions:
     code_samples:
   - name: os.clock
     summary: |
-      Returns a high-precision amount of CPU time used by Lua in seconds,
+      Returns a high-precision amount of elapsed CPU time in seconds,
       intended for use in benchmarking.
     description: |
-      Returns the amount of CPU time used by Lua in seconds. This value has high
+      Returns the amount of elapsed CPU time in seconds. This value has high
       precision, about 1 microsecond, and is intended for use in benchmarking.
 
       ```lua


content/en-us/reference/engine/libraries/os.yaml
Paracosm-Daemon, line 1: I think `amount of elapsed CPU time` a redundant statement. It could probably be shortened to just `elapsed CPU time`

content/en-us/reference/engine/libraries/os.yaml
```diff
@@ -300,10 +300,10 @@ functions:
     code_samples:
   - name: os.clock
     summary: |
-      Returns a high-precision amount of CPU time used by Lua in seconds,
+      Returns a high-precision amount of elapsed CPU time in seconds,
       intended for use in benchmarking.
     description: |
-      Returns the amount of CPU time used by Lua in seconds. This value has high
+      Returns the amount of elapsed CPU time in seconds. This value has high
```
Paracosm-Daemon, line 306: I think `amount of elapsed CPU time` a redundant statement. It could probably be shortened to just `elapsed CPU time`

content/en-us/reference/engine/libraries/os.yaml
```diff
@@ -300,10 +300,10 @@ functions:
     code_samples:
   - name: os.clock
     summary: |
-      Returns a high-precision amount of CPU time used by Lua in seconds,
+      Returns elapsed CPU time in seconds with high precision,
       intended for use in benchmarking.
     description: |
-      Returns the amount of CPU time used by Lua in seconds. This value has high
+      Returns elapsed CPU time in seconds. This value has high
       precision, about 1 microsecond, and is intended for use in benchmarking.
```
IgnisRBX, line 307:
```suggestion
      precision (about 1 microsecond) and is intended for use in benchmarking.
```

IgnisRBX (approved):

PR #772, opened 2024-08-17
Improve wording for select()

Flikter:
## Changes

This pull request improves wording for the global function `select`. A sentence with the same meaning is also removed.

## Checks

### DIFF
--- a/content/en-us/reference/engine/globals/LuaGlobals.yaml
+++ b/content/en-us/reference/engine/globals/LuaGlobals.yaml
@@ -483,7 +483,7 @@ functions:
         default:
         summary: |
           The index of the argument to return all arguments after in `args`. If
-          it's set to "#", the number of arguments that were passed after it is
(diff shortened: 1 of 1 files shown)


content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -471,26 +471,20 @@ functions:
...
-      ```
     parameters:
       - name: index
         type: Variant
         default:
         summary: |
-          The index of the argument to return all arguments after in `args`. or
-          must be "#".
+          The index of the argument to return all arguments after in `args`. If
+          it's set to "#", the number of arguments that were passed after it is
```
IgnisRBX, line 486:
```suggestion
          it's set to `"#"`, the number of arguments that were passed after it is
```

IgnisRBX (approved):

PR #805, opened 2024-09-03
Update TeleportService.yaml

TheyCallMeRyan:
## Changes

The API documentation here acts as if TeleportAsyncResult will always be returned. It does not notify of the result that users may experience if TeleportOptions is not passed. As such, it should be denoted as an optional parameter and the result marked as an optional result.

See: https://devforum.roblox.com/t/teleportasync-not-returning-a-teleportasyncresult/3145215/13

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/TeleportService.yaml
+++ b/content/en-us/reference/engine/classes/TeleportService.yaml
@@ -1031,12 +1031,12 @@ methods:
         default: nil
         summary: |
           An optional `Class.TeleportOptions` object containing additional
-          arguments to the TeleportAsync call.
+          arguments to the TeleportAsync call. If this is not passed, no result will be returned.
     returns:
       - type: Instance
         summary: |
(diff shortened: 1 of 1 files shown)


content/en-us/reference/engine/classes/TeleportService.yaml
```diff
@@ -912,7 +647,7 @@ methods:
       instances associated with the server access code, the
       `Class.DataModel.JobId` is not.
 
-      #### Studio limitation
+      #### Studio Limitation
 
       This service does not work during playtesting in Roblox Studio &mdash; To
```
IgnisRBX, line 652:
```suggestion
      This service does not work during playtesting in Roblox Studio; to
```

IgnisRBX:
In addition to your suggestion, I've cleaned up the redundant and repetitive details on the former teleport functions, urging usage of `TeleportAsync()` instead for most cases.

IgnisRBX (approved):

PR #829, opened 2024-09-17
Pathfinding patch

metricrb:
## Changes

When we call Humanoid:MoveTo() from waypoint to waypoint, and use the MoveToFinished event to detect when the character reaches each waypoint it does work. However unlike the demo place and code portrayed in the documentation on Character Pathfinding, many games will not operate with smoothness due to the event having a delay on being communicated, leaving a stuttering NPC between nodes and overall a poor visual effect.

This is detrimental on considering performance in public games, unlike test baseplates with minimal events and memory.

To resolve this issue, I implemented a snippet of ClickToMoveController under Player > PlayerScripts > PlayerModule > ControlModule. 

You should notice that ClickToMove controller does not register for MoveToFinished but instead uses its own logic to decide when to move to next waypoint - namely the RunService.RenderStepped event. This would avoid waiting for the delayed event.

https://devforum.roblox.com/t/npc-moveto-has-a-stuttering-effect-whilst-using-pathfindingservice/3067241?u=m_etrics

## Checks

### DIFF
--- a/content/en-us/characters/pathfinding.md
+++ b/content/en-us/characters/pathfinding.md
@@ -257,7 +257,7 @@ end
 
 ### Path Movement
 
-Each waypoint consists of both a **position** (`Datatype.Vector3`) and **action** (`Enum.PathWaypointAction|PathWaypointAction`). To move a character containing a `Class.Humanoid`, like a typical Roblox character, the easiest way is to call `Class.Humanoid:MoveTo()` from waypoint to waypoint, using the `Class.Humanoid.MoveToFinished|MoveToFinished` event to detect when the character reaches each waypoint.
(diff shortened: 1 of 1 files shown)


metricrb:
Hi, @IgnisRBX, you are completely right. It seems that the example code I submitted had a small mistake caused by me when I was copying it out of my game systems to function as an example. Please try again with the most recent commit. Apologies for any inconvenience.

IgnisRBX:
Hi again @metricrb ,
I tested your updated code and it seems now that the character is attempting to move toward the final waypoint in the path, not the next one in the series of waypoints. Are you seeing the same behavior?

PR #847, opened 2024-09-28
New DeepClone code snippet

EnumEnv:
## Changes

I've made a bit more documented of a code snippet to help scripters navigate their way through it, considering not everyone viewing would be experienced. This take on the deep clone snippet also gives a faster result, meaning the code is ran faster and you get your value returned faster.

OLD: **0.0000051** (os.clock)
NEW: **0.0000034** (os.clock)

The difference may be small but definitely in some cases noticeable and useful. This is due to the usage of the generalized iterator. Also simplifying the code with ternary operators. But, I believe my take on this new snippet would definitely be faster, and also educate other scripters better, even spiking their curiosity in code simplification if not familiar with ternary operators. All in short, I believe It's a great take.

https://create.roblox.com/docs/luau/tables#deep-clones
https://devforum.roblox.com/t/ternary-like-operation-in-lua/834254

## Checks


metatablecat:
table.clone would like to have a word with you about performance

IgnisRBX:
Hi @EnumEnv , in your version, where does `MY_deepCopy` come into play?

EnumEnv:
> Hi @EnumEnv , in your version, where does `MY_deepCopy` come into play?
@IgnisRBX 
Oh, my bad, that's actually a test thing I had set up before for recursion, was using 2 functions for comparing and well forgot to rename. I'll just change it back to the function's name "deepCopy". That's my bad

metatablecat:
Using generalised iteration (`for k, v in o do`) and shorthand if statements is faster, my table.clone idea is a lot slower lol.

image

EnumEnv:
@metatablecat The reason the 'your solution' one is slower is because you used "pairs", which I initially didn't use at all, I used the generalized iteration. So, that's the reason of it slowing down. Generally it should be faster.

EnumEnv:
Waiting on another review. Tested everything and re-benchmarked.

content/en-us/luau/tables.md
```diff
@@ -291,14 +291,21 @@ local clone = table.clone(original)
 To copy a more complex table with nested tables inside it, you'll need to use a recursive function similar to the following:
 
 ```lua
-local function deepCopy(original)
+-- The function used for deep copying a table.
```
IgnisRBX, line 294:
```suggestion
-- The function used for deep copying a table
```

IgnisRBX:
@EnumEnv looks good now! I just did minor edits to the comment lines (no ending periods as per our typical style, etc.)

IgnisRBX (approved):

PR #876, opened 2024-10-20
Linkify CoreScript

ccuser44:
## Changes

Linkify the mention of CoreScripts in `Instance.yaml`.
Now this could also refer to other types of scripts with high security context levels but that really doesn't matter in regards to the docs.

## Checks


IgnisRBX:
There's no longer a page for CoreScript since it's internal-only, so this link will not resolve. Thanks for the suggestion but I'd prefer to keep it un-linked.

PR #941, opened 2024-12-10
Humanoid.yaml: Remove "Unequip" sentence in :EquipTool description.

TheWandererLee:
The sentence regarding unequipping tool does not apply to :EquipTool(). It applies to :UnequipTool()

## Changes

Remove the incorrectly placed sentence.

## Checks


content/en-us/reference/engine/classes/Humanoid.yaml
```diff
@@ -1482,9 +1482,6 @@ methods:
       end
       ```
 
-      When this function is called, the humanoid will automatically unequip any
-      `Class.Tool|Tools` that it currently has equipped
```
IgnisRBX, line 1486: I think what this is _attempting_ to say (although not super clearly) is that calling Equip will unequip any currently equipped tool, "replacing" it with the new tool being equipped. Does that match behavior in testing? I haven't tested tools in awhile personally, but I think equipping a tool automatically unequips a current one...

TheWandererLee: Ah, you're correct. That is its meaning. In testing I see that currently equipped tools are indeed unequipped when this function is called. Additionally, I also found that it is possible to equip multiple tools by parenting them to the player's character. So, in this case, even the plural wording of the initial sentence is correct.

I've revised my sentence. Alternatively, this PR could be closed, since the initial meaning is correct.

IgnisRBX (approved):

PR #984, opened 2025-01-12
Optimization of Collision Group Script

Lucatacos:
## Changes

I've optimized the script for handling collision groups in Roblox. The main change involves removing the connection to DescendantAdded, applying the collision group immediately to all parts in the model upon character addition. This reduces unnecessary event connections, improving performance by applying the collision group once for all descendants, rather than repeatedly when new parts are added. This should help improve game performance, especially when handling many parts or players.

## Checks

### DIFF
--- a/content/en-us/workspace/collisions.md
+++ b/content/en-us/workspace/collisions.md
@@ -267,22 +267,22 @@ PhysicsService:RegisterCollisionGroup(CollisionGroupName)
 PhysicsService:CollisionGroupSetCollidable(CollisionGroupName, CollisionGroupName, false)
 
 local function setCollisionGroup(model)
-    -- Apply collision group to all existing parts in the model
-    for _, descendant in pairs(model:GetDescendants()) do
-        if descendant:IsA("BasePart") then
-            descendant.CollisionGroup = CollisionGroupName
-        end
-    end
+	-- Apply collision group to all existing parts in the model
+	for _, descendant in model:GetDescendants() do
+		if descendant:IsA("BasePart") then
+			descendant.CollisionGroup = CollisionGroupName
+		end
+	end
 end
 
 Players.PlayerAdded:Connect(function(player)
-    player.CharacterAdded:Connect(function(character)
-        setCollisionGroup(character)
-    end)
-    -- If the player already has a character, apply the collision group immediately
-    if player.Character then
-        setCollisionGroup(player.Character)
-    end
+	player.CharacterAdded:Connect(function(character)
+		setCollisionGroup(character)
+	end)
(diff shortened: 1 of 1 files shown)


content/en-us/workspace/collisions.md
```diff
@@ -262,27 +262,27 @@ interesting but unintended gameplay, such as characters jumping on top of each o
...
+PhysicsService:RegisterCollisionGroup(CollisionGroupName)
+PhysicsService:CollisionGroupSetCollidable(CollisionGroupName, CollisionGroupName, false)
+
+local function setCollisionGroup(model)
+    -- Apply collision group to all existing parts in the model
+    for _, descendant in pairs(model:GetDescendants()) do
+        if descendant:IsA("BasePart") then
+            descendant.CollisionGroup = CollisionGroupName
+        end
+    end
```
IgnisRBX, line 275:
```suggestion
local function setCollisionGroup(model)
	-- Apply collision group to all existing parts in the model
	for _, descendant in model:GetDescendants() do
		if descendant:IsA("BasePart") then
			descendant.CollisionGroup = CollisionGroupName
		end
	end
```

content/en-us/workspace/collisions.md
```diff
@@ -262,27 +262,27 @@ interesting but unintended gameplay, such as characters jumping on top of each o
...
 Players.PlayerAdded:Connect(function(player)
-	-- Detect when the player's character is added
-	player.CharacterAdded:Connect(onCharacterAdded)
+    player.CharacterAdded:Connect(function(character)
+        setCollisionGroup(character)
+    end)
+    -- If the player already has a character, apply the collision group immediately
+    if player.Character then
+        setCollisionGroup(player.Character)
+    end
```
IgnisRBX, line 285:
```suggestion
	player.CharacterAdded:Connect(function(character)
		setCollisionGroup(character)
	end)
	-- If the player already has a character, apply the collision group immediately
	if player.Character then
		setCollisionGroup(player.Character)
	end
```

IgnisRBX:
I was initially a bit concerned by this change, in that all of the character BaseParts wouldn't be loaded in by the time CharacterAdded fires, but all of the collision-prone parts should be loaded at this point. If you detect otherwise in any future testing, please file another PR to restore the DescendantAdded bit which would account for any BaseParts added after the CharacterAdded event. Thanks!

IgnisRBX (approved):

IgnisRBX:
FYI, I also removed the `pairs()` bit of the loop, since you can now omit `pairs()` and just operate on the table directly. :)

PR #1015, opened 2025-02-06
Update RaycastResult.yaml

lack-of-gravity-jack:
## Changes

Changed the references from "instance" to BasePart. Referring to the type of the intersected object as "instance" is too generic and may be confusing, since Models and Folders are also instances that some intersected object could have been a child of, but raycasts can only return a reference to the object in of itself and never the parent model. The (poorly named) _instance_ property of RaycastResult can only ever hold a basepart or a terrain cell, and terrain descends from basepart so basepart is accurate and more specific than instance.

## Checks

- [x ] This contribution was created in whole or in part by me, and I have the right to submit it under the terms of this repository's open source licenses.
- [x ] I understand and agree that this contribution and a record of it are public, maintained indefinitely, and may be redistributed under the terms of this repository's open source licenses.
- [x ] To the best of my knowledge, all proposed changes are accurate.


content/en-us/reference/engine/datatypes/RaycastResult.yaml
```diff
@@ -50,7 +50,7 @@ properties:
       The position of the intersection between the ray and the part.
```
IgnisRBX, line 50:
```suggestion
      The world space point at which the intersection occurred.
```

IgnisRBX: Updating this to read more like the description instead of generic and somewhat vague "part". :)

content/en-us/reference/engine/datatypes/RaycastResult.yaml
```diff
@@ -50,7 +50,7 @@ properties:
...
     description: |
       The world space point at which the intersection occurred, usually a point
-      directly on the surface of the instance.
+      directly on the surface of the `Class.BasePart`.
```
IgnisRBX, line 53:
```suggestion
      directly on the surface of the `Class.BasePart` or `Class.Terrain` cell.
```

IgnisRBX: Just adding this for clarity, since (absolutely) Terrain inherits from BasePart, but some creators don't realize that and this adds some clarity for them.

IgnisRBX (approved):

PR #1023, opened 2025-02-12
Clarify HapticEffect is not out on HapticService deprecation msg

dgxo:
## Changes

Change deprecation notice in HapticService to clarify that HapticEffect is not yet available, and set an alias so `Class.` is not visible but it will still link to `HapticEffect` when it is added.

Report: https://devforum.roblox.com/t/feedback-on-hapticservice/3466028/2
## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/HapticService.yaml
+++ b/content/en-us/reference/engine/classes/HapticService.yaml
@@ -24,7 +24,9 @@ tags:
   - Service
   - NotReplicated
 deprecation_message: |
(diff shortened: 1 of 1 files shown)


dgxo:
Perhaps the text inside the deprecation notice could be moved into the description and reworded, to let developers know that a replacement is coming?

sammygrey:
Change not needed:
The HapticService class is still deprecated regardless of the current implementation status of HapticEffect. HapticEffect is the intended future replacement for HapticService and even though it isn't yet implemented or has documentation, It's a nice feature for users looking at the HapticService page.

dgxo:
right whats prettier done here then

IgnisRBX (approved):

PR #1036, opened 2025-02-26
Clarify how AnimationTrack:Stop() with a fadeTime of 0 affects Motor6Ds

Starveldt:
## Changes

When calling `AnimationTrack:Stop()` with a fadeTime of 0, poses do not freeze in place by intention. It happens due to the `MaxVelocity` of Motor6Ds in the rig being zero, meaning they cannot return back to their desired angle (since they inherit Motors).

I believe the misconception in the passage I edited was due to the default StarterCharacter rigs having the MaxVelocity of all joints set to zero rather than something like TAU (which should be optimal, given MaxVelocity is in radians). I've also tested this below:

Rig with MaxVelocity set to 0 for all Motor6Ds, calling `AnimationTrack:Stop(0)`:
RobloxStudioBeta_aAWKhiW44t

Rig with MaxVelocity set to TAU for all Motor6Ds, calling `AnimationTrack:Stop(0)`:
RobloxStudioBeta_erHkiFJMix

Bones are not affected by this issue since they do not inherit Motor.

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/AnimationTrack.yaml
+++ b/content/en-us/reference/engine/classes/AnimationTrack.yaml
@@ -617,8 +617,11 @@ methods:
       of the initial weight of the animation.
 
       It is not recommended to use a fadeTime of 0 seconds to try to override
-      this effect and end the animation immediately as presently, this causes
-      the `Class.AnimationTrack` poses to freeze.
+      this effect and end the animation immediately for `Class.Motor6D|Motor6Ds`
+      that have their `Class.Motor.MaxVelocity` set to zero, as this causes
+      the joints to freeze in place. If it must end immediately, ensure the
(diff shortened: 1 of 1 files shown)


content/en-us/reference/engine/classes/AnimationTrack.yaml
```diff
@@ -316,7 +316,7 @@ properties:
     description: |
       AnimationTrack.WeightTarget is a read-only property that gives the current
       weight of the `Class.AnimationTrack`. It has a default value of 1 and is
-      set when `Class.AnimationTrack:Play()`, `Class.AnimationTrack:Stop()` or
+      set when `Class.AnimationTrack:Play()`, `Class.AnimationTrack:()` or
```
IgnisRBX, line 319:
```suggestion
      set when `Class.AnimationTrack:Play()`, `Class.AnimationTrack:Stop()` or
```

PR #1049, opened 2025-03-12
Improve documentation for CreateAssetAsync

metatablecat:
## Changes

The asset creation currently is only usable by plugins loaded in the LocalPlugin folder, this change marks that in the two relevant methods.

(possible TODO: moderation note?)

## Checks

### DIFF
--- a/content/en-us/reference/engine/classes/AssetService.yaml
+++ b/content/en-us/reference/engine/classes/AssetService.yaml
@@ -75,10 +75,12 @@ methods:
     writeCapabilities: []
   - name: AssetService:CreateAssetAsync
     summary: |
-      Creates a new asset from the given object.
+      Uploads a new asset to Roblox from the given object.
     description: |
-      Creates a new asset from the given object in Plugin or Open Cloud Luau
-      Execution context.
+      Uploads a new asset to Roblox from the given object.
+
+      This method can only be used in locally loaded plugins and uploads
+      assets without prompting first.
     code_samples:
       - AssetService-CreateAssetAsync
     parameters:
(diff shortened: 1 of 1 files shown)


metatablecat:
Scrapped the TODO, ready for review.

content/en-us/reference/engine/classes/AssetService.yaml
```diff
@@ -133,8 +135,10 @@ methods:
     summary: |
       Creates a new version for an existing asset from the given object.
```
itsfrank, line 136: Use Upload here too for consistency?
```suggestion
      Uploads a new version for an existing asset from the given object.
```

content/en-us/reference/engine/classes/AssetService.yaml
```diff
@@ -75,10 +75,12 @@ methods:
...
     summary: |
-      Creates a new asset from the given object.
+      Uploads a new asset to Roblox from the given object.
     description: |
-      Creates a new asset from the given object in Plugin or Open Cloud Luau
-      Execution context.
+      Uploads a new asset to Roblox from the given object.
+
+      This method can only be used in locally loaded plugins and uploads
+      assets without prompting first.
```
itsfrank, line 83: I think this is valuable information, i would like to add that this is temporary (even though it has taken much longer than I would have like to get support for published plugins :( ).

I'm not too familiar with our docs standards, does something like this make sense?
```suggestion
      (Temporary) For now, this method can only be used in locally loaded plugins and uploads
      assets without prompting first.
```

content/en-us/reference/engine/classes/AssetService.yaml
```diff
@@ -133,8 +135,10 @@ methods:
     summary: |
       Creates a new version for an existing asset from the given object.
     description: |
-      Creates a new version for an existing asset from the given object in
-      Plugin or Open Cloud Luau Execution context.
+      Uploads a new version for an existing asset from the given object.
+
+      This method can only be used in locally loaded plugins and uploads
+      assets without prompting first.
```
itsfrank, line 141: If we go with the "temporary" wording, we should also add it here

itsfrank (commented):
Thank you for taking the time to improve the docs, I think this is definitely an improvement, I have some minor suggestions. I'm not extremely familiar with our docs language, so feel free to point out if my suggestions are inappropriate.

IgnisRBX (approved):

PR #1053, opened 2025-03-18
Added engine detail to Camera.yaml

C-Corpze:
A detail about camera behavior such as cameras getting parented to `nil` when `Workspace.CurrentCamera` is set to another camera.

Current Roblox documentation lacks information about what happens when multiple camera objects are present in Workspace and the CurrentCamera property is changed to a different camera.

Cameras randomly being parented to `nil` might seem confusing and like some kind of bug if you can't even find anything about it in the documentation.


IgnisRBX:
Hi @C-Corpze ,
Good point to add, thanks! I assume that "discarded" camera will follow typical garbage-collection behavior and actually get destroyed unless you're holding on to them through a script or similar? In other words, one `Workspace.CurrentCamera` is changed and the other Cameras are parented to `nil`, I assume that they are eventually garbage-collected and destroyed unless there's some intentional script or link that's keeping them around?

C-Corpze:
I'm uncertain if cameras are GC'd when parented to nil, not entirely sure how to test this again.

What I did was create a bunch of cameras and whenever something changed, I would print their parent and attempt to reparent them back into workspace.

I found out about this weird behavior when I was coding my own camera system from scratch, where I intend to have multiple camera objects that each can have their own independent post-processing effects and whatnot.

I honestly wish this was customizable as I find it quite inconvenient at best that this keeps happening, I cannot rely on cameras always existing and instead have to implement some additional code checks to make sure I'm not trying to set the CFrame of something that is nil.

ExtremePro11299 (approved):

IgnisRBX:
Hi again @C-Corpze ,
Curious, how and when are you creating the additional cameras? I did a bit of testing and additional cameras don't seem to replicate at all once play begins, even if they're parented somewhere outside of Workspace. Are you creating them in a script after play begins?

C-Corpze:
> Hi again @C-Corpze , Curious, how and when are you creating the additional cameras? I did a bit of testing and additional cameras don't seem to replicate at all once play begins, even if they're parented somewhere outside of Workspace. Are you creating them in a script after play begins?

To clarify, I simply create camera objects on the client using a local script and then parent them to workspace.
From the moment CurrentCamera is changed, the previous camera will be parented to `nil`.
This behavior is consistent and keeps happening.

I assume the same thing happens on the server as well when it's done in a server script.
Although, typically you don't create cameras on the server so this is of course not tested yet.

content/en-us/reference/engine/classes/Camera.yaml
```diff
@@ -31,6 +31,15 @@ description: |
...
   information on how to adjust and customize the camera's behavior.
+
+
+  Warning:
+  It should also be noted that when changing `Workspace.CurrentCamera` to a new `Camera`
+  and there is more than a single `Camera` present in the top-hierarchy of `Workspace`
+  then all other unused `Camera` objects shall be parented to `nil` but not `Destroy()`d.
+
+  If you require a `Camera` to stay around, it is recommended to parent it inside a `Model` or object other than directly under `Workspace`.
+
```
IgnisRBX, line 42:
```suggestion

  #### Storing Multiple Cameras
  
  Note that when changing `Class.Workspace.CurrentCamera` to a new `Class.Camera`,
  all other `Class.Camera|Cameras` directly descending from `Class.Workspace` will be
  destroyed. If you need to store multiple cameras and swap between them on demand,
  it's recommended that you store them in a `Class.Folder` or `Class.Model` under
  `Class.Workspace`, inside which they will remain even when
  `Class.Workspace.CurrentCamera|CurrentCamera` is changed.
```

IgnisRBX (approved):

IgnisRBX (approved):

PR #1073, opened 2025-04-12
Added Next Gen Studio Preview note to PluginToolbarButton

ExtremePro11299:
## Changes

When I enabled the Next Gen Studio Preview option in the beta options, all plugin buttons were clickable when the viewport was hidden, so I mentioned this in the docs.

## Checks


IgnisRBX:
Hi @ExtremePro11299 ,
Thanks for the docs update request. This behavior will eventually be synced to work the same as it currently does, so I'm going to close this request since it would cite a behavior that currently works one way (now) but should work another way (as stated) and that will be fixed.

PR #1079, opened 2025-04-21
Update RunService.yaml

ColRealPro:
## Changes

I have changed the description of the `DeltaTimeSim` param to correctly say it is the time since the last simulation step instead of the last frame

```lua
local dt1 = game:GetService("RunService").PreSimulation:Wait()
local dt2 = game:GetService("RunService").PreSimulation:Wait() 
print(dt1, dt2) -- almost always outputs the same values, the physics simulation runs at a fixed 240hz afaik which would result in this
```

## Checks


IgnisRBX:
Hi @ColRealPro Could you please explain in more detail why changing it to "simulation step" is more accurate than "frame"? All of the other events for RunService are stated in "frame" terms, so curious why this should be an exception...

ColRealPro:
Hi @IgnisRBX I think that simulation step makes more sense for this because unlike the other events in RunService, this one's deltaTime isn't actually based on the time since the last frame, but rather is always in increments of 1/240 (0.00416, 0.00833, 0.0125) since the physics engine does steps of 1/240, this also means that if your fps is above 240 that PreSimulation may not run every frame as the docs say
image

Technically it's not actually the time since the last step exactly, maybe it should be clarified that it's usually in increments of 1/240?

IgnisRBX:
Hi again @ColRealPro ,
I made some further edits based on engineer feedback, including updates to `PostSimulation` and `PreAnimation`.

ColRealPro:
Hello again @IgnisRBX, I think these changes are great! LGTM

IgnisRBX (dismissed):

IgnisRBX:
Hi @ColRealPro ,
Just FYI, I also became aware of some confusion that RenderStepped and Stepped were marked as "deprecated" when, in fact, they are not (however they have been superseded by newer events). So, I removed their deprecation notes in this PR as well, and noted that they're just superseded.

IgnisRBX (approved):

PR #1085, opened 2025-04-26
Removed duplicates of debug.info()

ExtremePro11299:
## Changes

I ran into the debug.info() thing 3 times while scrolling. This should fix it. One of the duplicates had a different code sample that the other two, so I cut and pasted it into the one debug.info() that I made.

## Checks

### DIFF
--- a/content/en-us/reference/engine/libraries/debug.yaml
+++ b/content/en-us/reference/engine/libraries/debug.yaml
@@ -139,6 +139,20 @@ functions:
       fnA()
       ```
 
+      ```lua
+      local function fnA()
+
+      end
+
+      local function fnB()
+
+      end
+
+      -- Output line ("l"), name ("n"), and identifier ("f") for both fnA() and fnB()
+      print(debug.info(fnA, "lnf"))  --> 1 fnA function: 0x75e3d3c398a81252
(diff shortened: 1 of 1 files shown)


IgnisRBX:
Admitted that this is a bit unclear, but it's a factor of how library functions are output to docs. The three different "variations" of `debug.info()` are shown since they each expect a different set of parameters... note that it's either `level`, `options` OR `function`, `options` OR `thread`, `level`, `options`. So each of these is documented separately. Open to cleaning up aspects of each if necessary, but we'd prefer to keep all 3 variations.

ExtremePro11299:
Anti-stale comment

IgnisRBX:
How does this look? I basically consolidated the three `debug.info()` blocks into one, and noted that it can be done in three different ways. Did the same for the two `debug.traceback()` blocks as well.
image

IgnisRBX (approved):

tchu-roblox:
Hi @ExtremePro11299, thank you for your contribution! I appreciate the effort to clean up duplicate entries — the intent makes perfect sense.

Unfortunately, we'll need to revert this change for now. The current system relies on those duplicates to render documentation for overloaded functions in the Script Editor. We know this isn't ideal, and we're looking into ways to improve how these duplicates are navigated and displayed on the docs site.

Thanks again for taking the time to improve the docs — we value your contributions!

PR #1088, opened 2025-04-28
Update ProximityPromptService.yaml

incetents:
Adding more context to PromptTriggered working differently when used on the client.
If created on the Server, this event will resolve when any player interacts with the prompt.
If created on the Client, only the player who interacts with the prompt will have this event resolved.


IgnisRBX:
OK with adding these details, but the same policy apply to the other events of ProximityPromptService? If so, best to add it for them too. :)

PR #1114, opened 2025-05-25
Add example usage function for pcall()

sammygrey:
## Changes

The documentation can be somewhat confusing when trying to learn how to correctly implement pcalls as pointed out by @junveld in [this forum post](https://devforum.roblox.com/t/feedback-on-luau-globals/3664364). I have added a small piece of example code for how to use pcalls to do function calls that even when returning an error, will allow the rest of the script to continue as normal.

## Checks


content/en-us/reference/engine/globals/LuaGlobals.yaml
```diff
@@ -346,6 +346,20 @@ functions:
...
+      end
+      
+      local success, response = pcall(divideByFive, "notANumber") -- results in error
+      
+      if success then
+        -- handle successful response ...
+      else
+        warn("Error message: " .. response)
+      end
+      ```
```
IgnisRBX:
```suggestion
      ```lua
      local function divideByFive(n: number): number
      	return n / 5
      end
      
      local success, errorMessage = pcall(divideByFive, "notANumber")  -- Results in error...
      
      if success then
      	-- Handle successful response...
      else
      	warn("Error message:", errorMessage)
      end
      ```
```

IgnisRBX:
@sammygrey 
Thanks for the contribution! I just changed the pcall() param name from `response` to `errorMessage` to match what's generally used throughout our docs.

PR #1232, opened 2025-08-31
Expand and clarify the `__mode` metaproperty

Aseminaunz:
## Changes

Clarified and expanded the use of the `__mode` metaproperty.

The current description makes it seem to me as if a reference to a Roblox instance is ALWAYS strong, which is not true. This is only true if the instance isn't parented to `nil` as there is then a reference to it in the datamodel. Weak tables containing Roblox instances are subject to garbage collection, as per my own test:
```luau
local gcForcer = setmetatable({}, { __mode = "v" })

for i = 1, 10000 do
	gcForcer[i .. "buffer"] = buffer.create(50000)
	gcForcer[i .. "rbx"] = Instance.new("Part")
end

print("Done!")
task.wait(2)

local gcedBuffers = 0
local gcedInstances = 0

for i = 1, 1000 do
	if not gcForcer[i .. "buffer"] then
		gcedBuffers += 1
	end
	if not gcForcer[i .. "rbx"] then
		gcedInstances += 1
	end
end

print("GC'ed buffers", gcedBuffers)
print("GC'ed instances", gcedInstances)
-- All show up as garbage collected
```

Not all userdata, however, are subject to garbage collection, as duly noted. For example, Vector3:

```luau
local gcForcer = setmetatable({}, { __mode = "v" })

for i = 1, 10000 do
	gcForcer[i .. "buffer"] = buffer.create(50000)
	gcForcer[i .. "vec"] = Vector3.new(math.random())
end

print("Done!")
task.wait(2)

local gcedBuffers = 0
local gcedVectors = 0

for i = 1, 1000 do
	if not gcForcer[i .. "buffer"] then
		gcedBuffers += 1
	end
	if not gcForcer[i .. "vec"] then
		gcedVectors += 1
	end
end

print("GC'ed buffers", gcedBuffers)
print("GC'ed vectors", gcedVectors)
-- All buffers collected, no Vector3s are gone
```

This expanded explanation also clarifies what "weak" means, and links to the Lua manual for a more in-depth explanation.

## Checks


Aimarekin:
"Not all userdata" being garbage collectible might be a mistake now that I've learned Vector3.new simply returns a `vector` type. Some internal feedback on whether if all `userdata` is subject to garbage collection (which I assume they are, unless the engine keeps a reference) or not would be nice before I change that statement.

Aseminaunz:
"Not all userdata" being garbage collectible might be a mistake now that I've learned Vector3.new simply returns a `vector` type. Some internal feedback on whether if all `userdata` is subject to garbage collection (which I assume they are, unless the engine keeps a reference) or not would be nice before I change that statement.

content/en-us/luau/metatables.md
```diff
@@ -111,7 +111,7 @@ Here's the list of available metamethods:
 		</tr>
 	  <tr>
 		  <td>`__mode`</td>
-		  <td>Used in weak tables, declaring whether the keys and/or values of a table are weak. Note that references to Roblox instances are never weak. Tables that hold such references will never be garbage collected.</td>
+		  <td>A string making either all of the table's keys `"k"`, values `"v"`, or both `"kv"` weak. This means that the [garbage collector](https://www.lua.org/pil/17.html) will not count them as references, allowing them to be freed from memory when they're no longer used anywhere else. When a table entry's either key or value (as described by the string) is garbage collected, it will be removed (since one them is `nil`, or possibly both if using `"kv"`, the entry will no longer exist). Only tables, functions, `Library.buffer`s, and some [userdata](./userdata.md) are garbage collectable, other [types](./index.md#types) are not and won't be affected. Note that even if you don't explicitly hold a reference to an object in your code, the engine might still be keeping one (for example, if a Roblox `Instance` is a descendant of the `Datamodel`).</td>
```
IgnisRBX:
```suggestion
		  <td>A string making either all of the table's keys (`"k"`), values (`"v"`), or both (`"kv"`) weak. This means that the garbage collector will not count them as references, allowing them to be freed from memory when they're no longer used anywhere else. When a table entry's key or value (as described by the string) is garbage collected, it will be removed since one them is `nil` (or possibly both if using `"kv"`) and the entry will no longer exist. Only tables, functions, buffers (`Library.buffer`), and some [userdata](./userdata.md) are garbage collectable; other [types](./index.md#types) are not and won't be affected. Note that even if you don't explicitly hold a reference to an object in your code, the engine might still be keeping one, for example if a Roblox `Class.Instance` is a descendant of the `Class.DataModel`.</td>
```

vegorov-rbx:
All userdata types as well as well as thread objects (from coroutine and task libraries) are garbage-collectable and can be weak.
Vector3 is a primitive type instead of userdata now.

One other exception are string values.
Quoting Lua 5.3 manual (Luau is based on 5.1, but manual for that did not specify this):
> Although strings are subject to garbage collection, they do not have an explicit construction, and therefore are not removed from weak tables.

PR #1250, opened 2025-09-17
Remove Enum from Luau Reference section

gaymeowing:
Moved the Enum page to be under the Scripting section instead of Luau Reference.
Also fixed the luau url on the index page for the scripting Section, and added a warning about using the `Value` property of EnumItems.


IgnisRBX:
Hi @gaymeowing ,
Is your suggestion to also move Enums up into the main Scripting branch in the left-side TOC? I think that would look a bit misplaced because it would then sit beside other higher-level scripting articles.

image

Admittedly, Enums are not strictly a "Luau type" so the current placement isn't perfect, but it does sit beside other "types you can assign to variables or use in conditional statements" just like nil, booleans, strings, etc. in that TOC branch, so it's fairly logical IMO. If you have other suggestions on how to reorganize the current articles in the Scripting branch, please comment here and we can look into it.

gaymeowing:
Perhaps a DataTypes section under Scripting? As there's also an article on CFrames, and Content (bit odd but I think the docs on Content paths like `rbxthumb://` and the Content DataType itself could be separated out)

aetter:
Hey all, yeah, rather than moving anything, I think it's worth just noting on the enums page, "These aren't technically a Luau type, just Engine presets that you access through the enum data type, but they're conceptually similar enough and something you work with all the time in Roblox development." The other option would be to just delete the page, but the content is decent enough that I think the lesser evil is to leave it somewhere not-quite-logically consistent.

IgnisRBX:
Hi @gaymeowing ,
Could you please revert the file location of the Enums page in this PR? Then I can proceed to review the other changes and merge it in. Thanks!

content/en-us/luau/enums.md
```diff
@@ -3,6 +3,10 @@ title: Enums
 description: A fixed list of items (enumeration).
 ---
 
+<Alert severity="info">
+Enums are not a [builtin luau type](https://luau.org/typecheck#builtin-types), and exist only in Roblox. This page is located here because Eunms are conceptually similar enough, and are something you will work with all the time in Roblox development.
+</Alert>
```
IgnisRBX, line 8:
```suggestion
<Alert severity="info">
Enums are not a [built-in Luau type](https://luau.org/typecheck#builtin-types) and they exist only in Roblox, but they're conceptually similar to other Luau data types and are something you'll work with frequently in Roblox development.
</Alert>
```

content/en-us/luau/enums.md
```diff
@@ -3,6 +3,10 @@ title: Enums
...
+
 The **enumeration** data type, or `Datatype.Enum`, is a fixed list of items. You can access enums through the global object called `Datatype.Enum`. For a full list of Enums and their items, see [Enums](/reference/engine/enums) in the API Reference.
```
IgnisRBX:
```suggestion
The **enumeration** data type, or `Datatype.Enum`, is a fixed list of items. You can access enums through the global object called `Datatype.Enum`. For a full list and their respective items, see [Enums](/reference/engine/enums).
```

content/en-us/luau/enums.md
```diff
@@ -3,6 +3,10 @@ title: Enums
...
 
 ## Get enum items
```
IgnisRBX:
```suggestion
## Enum items
```

content/en-us/luau/enums.md
```diff
@@ -3,14 +3,19 @@ title: Enums
...
+<Alert severity="info">
+Enums are not a [built-in Luau type](https://luau.org/typecheck#builtin-types) and they exist only in Roblox, but they're conceptually similar to other Luau data types and are something you'll work with frequently in Roblox development.
+</Alert>
 
-## Get enum items
+The **enumeration** data type, or `Datatype.Enum`, is a fixed list of items. You can access enums through the global object called `Datatype.Enum`. For a full list and their respective items, see [Enums](/reference/engine/enums).
+
+## Enum items
 
 To get all items of an Enum, call the `GetEnumItems()` method on the enum. The following code sample demonstrates how to call `GetEnumItems()` on the `Enum.PartType` enum.
```
IgnisRBX:
```suggestion
To get all items of an enum, call the `Datatype.Enum:GetEnumItems()|GetEnumItems()` method on it. The following code sample demonstrates how to call `Datatype.Enum:GetEnumItems()|GetEnumItems()` on the `Enum.PartType` enum.
```

content/en-us/luau/enums.md
```diff
@@ -19,30 +24,34 @@ end
...
+print(Enum.PartType.Cylinder.Name) --> "Cylinder"
+print(Enum.PartType.Cylinder.Value) --> 2
+print(Enum.PartType.Cylinder.EnumType) --> PartType
 ```
 
-## Assign enum items
-
 To assign an `Datatype.EnumItem` as the value of a property, use the full `Datatype.Enum` declaration. You can also use its `Value` or `EnumType`.
 
+<Alert severity="warning">
```
IgnisRBX: @gaymeowing ,
Just curious, can you provide an example of when this happened and broke code? When an enum item is added to an enum, the previous items and values should remain as they always were... meaning, the addition of a new item should not "re-sort" the value integers on an enum, but rather a new item/value is added.

That being said, we can probably just remove mention of assigning enum "values" (integers) as properties from this guide, even though it's possible to do so. IMHO the values are largely meaningless for anybody else reading the code and I'd be OK with just omitting it from our guidance.

gaymeowing: No I've just heard of it happening before iirc, but emitting sounds good. 
Although emitting being able to assign strings also seems like a good idea? As that's not really good practice either

IgnisRBX (approved):

brionawilliams2026-web:
hi

PR #1444, opened 2026-02-17
Please change this argument name "contentIdList" to something so it doesn't sound like direct ID input

karl-police:
Please change this argument name "contentIdList" to something so it doesn't sound like direct ID input, array of string of rbxassetid please. I don't know, but I think there used to be an actual method that accepted a string directly, while this one is just instances which can hold ids.

Or the API itself doesn't throw if no instance is passed.

also seeking answer for this still: https://github.com/Roblox/creator-docs/pull/1386#issue-3794263566

> I wanted to try to add other tags. But also Groups, but all in one page. The problem is this page is for tags only right now, which is sad.

and the other one was about the chat update being clarified in the docs, but I think that's confirmed to be WIP all over around the place


michaeltipman-gif:
No cussing

karl-police:
uhh

danelynberuela-rgb (commented):
1444

aetter:
Hey @karl-police I'm going to close this out, thanks as always. Descriptions and summaries are simple enough, but parameter names come from further upstream and aren't something we can take external PRs on.

PR #1453, opened 2026-02-27
Start Documenting Roblox's Standard Scripts

karl-police:
## Checks

## Changes

The changes introduce early documentation about Roblox's Standard Scripts. This is important because things like the ``CameraModule`` do things is important to know about. For instance, the RenderPriority of the Camera and how it changes the ``CFrame`` of the camera. Otherwise, it causes confusion.

image

The current text and information is through careful analyzing and from Roblox's own comments within the code. There is not much documentation yet, because I happen to find bugs that need to be resolved before this documentation can continue. Until then, I want to share my vision on how these documentation could be structured and look like and hope to have this PR accepted, which provide this initial set up for documenting Roblox's Standard Scripts, as I did all these changes just by using the GitHub websites. I even had to temporarily change ``.gitattributes`` just to upload a PNG.

It also adds info about ``BoundKeys`` which is a ``StringValue`` that is comma-separated and can be overwritten to change the bindings for Shift-Lock.

This documentation is necessary so that there's also official guidelines on what can be done and what shouldn't be done. Keep in mind that the PlayerModule says that a Creator is allowed to access things that it provides, but never had a documentation.

Bug Reports:
- https://devforum.roblox.com/t/robloxs-cameramodule-doesnt-ever-return-the-camera-to-the-playermodule/4451947


karl-police (commented):
test

michaeltipman-gif:
No cousin

IgnisRBX:
Hi @karl-police ,
Thanks for the initial submission. I'm a bit hesitant on how much these scripts should be officially documented, honestly. I know that many creators clone them from a running game and modify them, and perhaps that's a great approach for their situation, but it's also a fairly "fragile" approach since these scripts change over time (without notice) and so the clones must be kept in continual sync with the new core changes and creator-applied modifications.

Are there specific points that you'd like clarified _from_ these scripts, such as camera rendering details that may not be well documented (or perhaps not at all) under related guides and API docs for Camera? If so, perhaps we can work in those details to other places...

karl-police:
> Hi @karl-police , Thanks for the initial submission. I'm a bit hesitant on how much these scripts should be officially documented, honestly. I know that many creators clone them from a running game and modify them, and perhaps that's a great approach for their situation, but it's also a fairly "fragile" approach since these scripts change over time (without notice) and so the clones must be kept in continual sync with the new core changes and creator-applied modifications.
> 
> Are there specific points that you'd like clarified _from_ these scripts, such as camera rendering details that may not be well documented (or perhaps not at all) under related guides and API docs for Camera? If so, perhaps we can work in those details to other places...

The camera rendering. It could go to RenderStepped and BindToRenderStep to give the clue on what to watch out for.

e.g.

- BindRenderStep 1# Changes Camera CFrame
- BindRenderStep 2# Fires raycast from camera
- BindRenderStep 3# Changes Camera CFrame again

- RenderStep RNG <-- this render step initialized first in the code
- RenderStep RNG 2# <-- this render step may also initialize first

One of these RenderStep fires a cast, while the other one changes the camera. If the cast happens before the camera change, you may get a unexpected result. But "RNG" because it's a race condition on which one "Connects" first.

Something like that.


&nbsp;


For the other things. The ObjectValues where you can change the keybinds, I think these are super specific to the Roblox Standard Scripts, yet useful. Since these scripts change overtime, maybe older documentation about them could get archived 🤷 but still be available in the Creator Docs

karl-police:
Found this old page: <https://web.archive.org/web/20200304050359/https://developer.roblox.com/en-us/articles/customizing-game-controls>



There's also a few other pages that I used to know about once. Like this one: <https://web.archive.org/web/20191014231924/https://developer.roblox.com/en-us/articles/Bezier-curves>
<https://web.archive.org/web/20230214174739/https://create.roblox.com/docs/mechanics/bezier-curves>

But I think there was another one with ease-in I think.


Not this one: <https://web.archive.org/web/20120803044539/http://wiki.roblox.com/index.php/Parametric_equations>

It was something else. It was like EasingStyle but with math instead.

Doesn't seem to be this one either <https://web.archive.org/web/20180805200548/http://robloxdev.com:80/articles/Lerp>

I didn't even know pages like these existed: <https://web.archive.org/web/20180207161209/http://wiki.roblox.com/index.php?title=Quaternion>

karl-police:
Also note

image

``FFlagUserPlayerModuleHiddenAPI``

Ransomwave:
> I'm a bit hesitant on how much these scripts should be officially documented, honestly. I know that many creators clone them from a running game and modify them, and perhaps that's a great approach for their situation, but it's also a fairly "fragile" approach since these scripts change over time (without notice) and so the clones must be kept in continual sync with the new core changes and creator-applied modifications.
> 
> Are there specific points that you'd like clarified _from_ these scripts, such as camera rendering details that may not be well documented (or perhaps not at all) under related guides and API docs for Camera? If so, perhaps we can work in those details to other places...

@IgnisRBX 

I understand the point you're making about "fragility", but I think that fragility is exactly why we need this documentation.

I heavily disagree with the idea that the Player scripts should be a "black box" that creators have to work around or against. It’s incredibly frustrating to have to reverse-engineer the architecture and intricacies of default behavior just to understand how it functions.

The reason creators clone and fork these scripts in the first place is specifically because there isn't a clear roadmap for extending them. When we're left guessing how these systems work at a "low level", we end up with the exact "fragile" forks you’re concerned about.

At the very minimum, there should be a high-level, bird's-eye view of what each script is responsible for, even if it includes a disclaimer that implementation details are subject to change.

IgnisRBX:
@karl-police 
Just FYI I'm looking into getting some documentation out for this. I won't recommend copying the scripts from runtime and using them as total overrides for the default scripts, but I'll look into some of the getter tactics you drafted up that don't involve the fragile cloning approach...

content/en-us/roblox-scripts/PlayerModule/CameraModule.md
```diff
@@ -0,0 +1,29 @@
...
+
+<i>The camera controller ModuleScripts implement classes which are instantiated and activated as-needed, they are no longer all instantiated up front as they were in the previous generation of PlayerScripts.</i>
+
+## cameraRenderUpdate
+
+``cameraRenderUpdate`` is the RenderStep name used and bound with by the ``CameraModule`` with the priority ``Enum.RenderPriority.Camera.Value``, through Class.RunService:BindToRenderStep()|:BindToRenderStep().
+
+For custom camera or other behavior, it is recommended to keep the **``Enum.RenderPriority``** in mind. The MicroProfiler can be used to debug which render steps are running in which order.
+
+The order is important. For instance, if the ``CameraModule`` changes the ``Datatype.CFrame`` of the camera,
```
IgnisRBX, line 16: @karl-police Can you explain more clearly what you mean here? Or show a code example of how you're changing `CFrames` and in which order you're doing so?

Are you basically saying to use `Enum.RenderPriority.Camera.Value - 1` in `BindToRenderStep()`?

IgnisRBX:
Hi again @karl-police ,
On further consideration, I'm opting to not document `PlayerModule`. Changes are always underway and we don't recommend accessing those modules or their getter functions (of course you _can_ do so, but with all the risk and fragility that doing so entails). If you think other changes in this PR should be applied elsewhere in docs, such as to the [Camera](https://create.roblox.com/docs/workspace/camera) guide, please open another PR and I'll take a look.

content/en-us/roblox-scripts/PlayerModule/CameraModule.md
karl-police: I'll put the reply below.

karl-police:
@IgnisRBX 

The thing I looked at was set up something like this

<details>

```luau
local Players = game:GetService("Players")
local RunService = game:GetService("RunService")

local player = Players.LocalPlayer
local camera = workspace.CurrentCamera

--camera.CameraType = Enum.CameraType.Scriptable

local CAMERA_OFFSET = Vector3.new(0, 60, 30)

-- after camera
--local renderPriority = Enum.RenderPriority.Camera.Value + 1

--[[RunService:BindToRenderStep("abc", renderPriority, function()
	local character = player.Character
	if character and character:FindFirstChild("HumanoidRootPart") then
		local rootPart = character.HumanoidRootPart
		local targetPos = rootPart.Position + CAMERA_OFFSET
		camera.CFrame = CFrame.new(targetPos, rootPart.Position)
	end
end)]]

RunService.RenderStepped:Connect(function()
	debug.profilebegin("Cam1")
	local character = player.Character
	if character and character:FindFirstChild("HumanoidRootPart") then
		local rootPart = character.HumanoidRootPart
		local targetPos = rootPart.Position + CAMERA_OFFSET
		camera.CFrame = CFrame.new(targetPos, rootPart.Position)
	end
	debug.profileend()
end)
```


```luau

local UserInputService = game:GetService("UserInputService")
local RunService = game:GetService("RunService")

local MAX_MOUSE_DISTANCE = 99999

local function getWorldMousePosition(raycastParams: RaycastParams)
	local mouseLocation = UserInputService:GetMouseLocation()

	-- Create a ray from the 2D mouseLocation
	local screenToWorldRay = workspace.CurrentCamera:ViewportPointToRay(mouseLocation.X, mouseLocation.Y)

	-- The unit direction vector of the ray multiplied by a maximum distance
	local directionVector = screenToWorldRay.Direction * MAX_MOUSE_DISTANCE

	-- Raycast from the ray's origin towards its direction
	local raycastResult = workspace:Raycast(screenToWorldRay.Origin, directionVector, raycastParams)

	if raycastResult then
		-- Return the 3D point of intersection
		return raycastResult.Position
	else
		-- No object was hit so calculate the position at the end of the ray
		return screenToWorldRay.Origin + directionVector
	end
end


local Part = Instance.new("Part")
Part.Anchored = true
Part.Parent = game.Workspace

local raycastParams = RaycastParams.new()
raycastParams.FilterType = Enum.RaycastFilterType.Exclude
raycastParams:AddToFilter(Part)

RunService:BindToRenderStep("testtest", Enum.RenderPriority.Last.Value, function(dT)
	debug.profilebegin("rayTest")
	local target = getWorldMousePosition(raycastParams)

	Part.CFrame = CFrame.new(target)
	debug.profileend()
end)
```

</details>


Now, I don't think the camera is involved here, but the camera was a sidethought.

So, for example, if you'd make this renderstep without ``.Scriptable`` happen before the camera ran, the camera would overwrite it.

Here in this case you will see that the cube is offset from the mouse, because of the render priority. If you set the "Cam1" example to ``Enum.RenderPriority.First.Value`` you'd see that you're now able to zoom in and out.

So, what it's basically saying is, if you're leaving the default camera script as is, and just making some kind of raycast, it should be made after the default camera has done things, so the offset is correct.

PR #1500, opened 2026-04-22
AncestryChanged clarification

karl-police:
## Changes

Changes the wording, as the current one is confusing, due to also the parameters being named ``child`` and ``parent``.

It clarifies a bit better that ``InstanceA.AncestryChanged``

That the parameter ``parent`` from the connection callback, is not ``InstanceA.parent``, but rather ``child.Parent``.

## Suggestion

I'd also like ``child`` and ``parent`` be renamed.

``childAncestor`` and ``newAncestorParent`` maybe.

or for other consistency

``childAncestorWhoChanged`` and ``newChildAncestorParent``


aetter:
Thanks for calling this out @karl-police. We generate the reference guides, so I'm going to close this out and make some slightly more extensive changes upstream.

PR #1545, opened 2026-07-22
Update oauth2-registration.md

T4CT1CALRANGER:
added another Info alert under the configure general information,  from most recent experience, I myself had issues with 3 false rejections, due to my URL's which were all Correct & functional, not being Specific enough, adding this will help others to not have to go searching to find why they got rejected when something like one of the specific Pages, was unable to be found..

## Changes

Added 2nd Alert to make sure people know that if their URL's aren't specific enough, they might get a false rejection.

## Checks

- [* ] This contribution was created in whole or in part by me, and I have the right to submit it under the terms of this repository's open source licenses.
- [* ] I understand and agree that this contribution and a record of it are public, maintained indefinitely, and may be redistributed under the terms of this repository's open source licenses.
- [ *] To the best of my knowledge, all proposed changes are accurate.

### DIFF
--- a/content/en-us/cloud/auth/oauth2-registration.md
+++ b/content/en-us/cloud/auth/oauth2-registration.md
@@ -44,6 +44,9 @@ Once you've [registered](#register-your-app) your app, you should configure its 
   You can skip this section during app development and testing, but you must
   complete it during the review process to publish your app to users.
 </Alert>
+<Alert severity="info">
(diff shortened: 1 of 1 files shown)


aetter:
Thanks for the PR @T4CT1CALRANGER but I'm going to close it out. It's an improvement we should make on our review end rather than a warning we should provide.

PR #1549, opened 2026-07-27
Clarify metatable usage and weak tables

soyuta:
## Changes

* Reorganized the use cases for metatables under clearer headings.
* Removed the ambiguous explanation of weak metatable behavior.
  The statement that "references to Roblox instances are never weak" could easily be misinterpreted to mean that references remain strong even inside weak tables. Likewise, readers might incorrectly assume that setting `__mode` causes an Instance to be collected automatically even when it is still alive in the game engine.
* Added use cases for weak metatables.
* Added a warning explaining that, when weakly referencing an Instance, the removal of all Luau references does not necessarily mean that the corresponding reference has also been released in the underlying C++ environment.

These behaviors have frequently been misunderstood. Many developers have assumed that an Instance stored in a weak table is collected only after the Instance itself has actually been destroyed.

As a result, some have also incorrectly reported the expected behavior as a bug when the Instance was not collected under those assumptions.

These sections should be clarified to prevent further confusion.

Related discussions:

* https://devforum.roblox.com/t/weak-tables-dont-work-with-instances/86634?u=sakura_soyuta
* https://devforum.roblox.com/t/studio-beta-reference-instances-directly-with-attributes/4753441/48?u=sakura_soyuta

## Checks

### DIFF
--- a/content/en-us/luau/metatables.md
+++ b/content/en-us/luau/metatables.md
@@ -111,7 +111,7 @@ Here's the list of available metamethods:
 		</tr>
 	  <tr>
 		  <td>`__mode`</td>
-		  <td>Used in weak tables, declaring whether the keys and/or values of a table are weak. Note that references to Roblox instances are never weak. Tables that hold such references will never be garbage collected.</td>
+		  <td>Used in weak tables, declaring whether the keys and/or values of a table are weak.</td>
 		</tr>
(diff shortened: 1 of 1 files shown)


aetter:
Thanks @soyuta. I'm going to merge and then make some additional tweaks internally.

